PR #27, opened 2019-03-20
[BTS-87] <Input />

ddsilva:
input


components/Input/Input.jsx
```diff
@@ -1,43 +1,35 @@
...
-  width: 100%;
+  margin-top: 8px;
+  padding-right: 42px;
+
+  ${DEFAULT_INPUT_STYLE};
+
+  ${({ searchable }) =>
+    searchable &&
+    `
+    padding: 10px 42px 10px 42px;
```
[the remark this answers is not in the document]
alan-oliv: or 
```suggestion
    padding: 10px 42px;
```

components/Input/Input.jsx
```diff
@@ -49,40 +41,77 @@ const InputTag = styled.input`
...
+  color: ${Colors.ERROR['500']};
 `;
 
 const InputErrorMessage = styled(ErrorMessage)`
-  margin-top: 8px;
+  padding: 8px 12px;
+`;
+
+const HelperText = styled.span`
+  color: ${Colors.BLACK['700']};
```
alan-oliv: In our conversations, we specified that the labels should inherit color from body, is that right?

components/Input/Input.jsx
```diff
@@ -49,40 +41,77 @@ const InputTag = styled.input`
...
+  cursor: text;
+  display: block;
+  font-size: 14px;
+  font-style: italic;
+  font-weight: 600;
+  padding: 8px 12px;
+`;
+
+const DescriptionLabel = styled.span`
+  color: ${Colors.BLACK['700']};
```
alan-oliv: "In our conversations, we specified that the labels should inherit color from body, is that right?"
this rule applies here to

PR #29, opened 2019-03-21
[BTS-168] TabbedView 

ggdaltoso:
image


components/TabbedView/TabbedView.jsx
```diff
@@ -0,0 +1,147 @@
...
+    },
+  };
+
+  return skins[skin] || skins.default;
+};
+
+const Navbar = styled.ul`
+  display: flex;
+  flex-grow: 1;
+  flex-shrink: 1;
```
allyssonsantos: I think those flex properties isn't necessary.

components/TabbedView/TabbedView.jsx
```diff
@@ -0,0 +1,147 @@
...
+  list-style: none;
+  margin: 0 0 25px 0;
+  padding: 0;
+`;
+
+Navbar.displayName = 'Navbar';
+
+const NavItem = styled.li`
```
allyssonsantos: Alphabetical order?

rapahaeru:
For a11y issues, can we use aria roles for this tabs?

For example:
```
<div class="tabs">
    <div role="tablist">
        <button role="tab" aria-selected="true" aria-controls="apple-panel" id="apple-tab">Apple</button>
        <button role="tab" aria-selected="false" aria-controls="orange-panel" id="orange-tab">Orange</button>
    </div>
    <div role="tabpanel" id="apple-panel" aria-labelledby="apple-tab">
        <p>Apple panel content</p>
    </div>
    <div role="tabpanel" id="orange-panel" aria-labelledby="orange-tab">
        <p>Orange panel content</p>
    </div>
</div>
```

Or should not we worry with this now?

components/TabbedView/TabbedView.jsx
```diff
@@ -0,0 +1,147 @@
...
+              skin={skin}
+            >
+              {title}
+            </NavItem>
+          ))}
+        </Navbar>
+
+        {React.Children.map(children, child => (
+          <RenderIf conditional={child.props.title === activeTab}>
+            {child.props.children}
```
rapahaeru: what do you think of having some another block element here?
I believe that if put one, makes easier to manipulate that element in the future

components/TabbedView/TabbedView.jsx
```diff
@@ -0,0 +1,147 @@
...
+  box-sizing: border-box;
+  cursor: pointer;
+  flex-shrink: 0;
+  font-size: 20px;
+  height: 48px;
+  line-height: 1.5;
+  min-width: 90px;
+  overflow: hidden;
+  padding: 9px 16px;
+  position: relative;
```
rapahaeru: do we need this property?

components/TabbedView/TabbedView.jsx
```diff
@@ -0,0 +1,147 @@
...
+  text-transform: uppercase;
+
+  ${({ skin, active }) => {
+    const { background, text, activeText, hover } = getColors(skin);
+    return css`
+      background-color: ${background};
+      color: ${text};
+
+      ${active &&
+        `
```
rapahaeru: when tha tab was active, we could let the cursor property with "default" instead "pointer".
Do you agree?

components/TabbedView/TabbedView.jsx
ggdaltoso: No and I must remove it. :+1:

PR #33, opened 2019-03-26
[BTS-179] Fix: Tooltip state management

alan-oliv:
Tooltip was causing maximum depth update exceeded, so we made some improvements.

### DIFF
--- a/components/Tooltip/Tooltip.unit.test.jsx
+++ b/components/Tooltip/Tooltip.unit.test.jsx
@@ -2,25 +2,43 @@ import React from 'react';
 import { mount } from 'enzyme';
 import Tooltip from './Tooltip';
 
+const TOOLTIP_TEXT = 'This is a hint';
+
 describe('Tooltip component ', () => {
   describe('All positions', () => {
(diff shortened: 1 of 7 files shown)


components/Tooltip/Tooltip.unit.test.jsx
```diff
@@ -2,25 +2,43 @@ import React from 'react';
...
 describe('Tooltip component ', () => {
   describe('All positions', () => {
     it('Should match the snapshot when place is top', () => {
-      const tooltip = mount(<Tooltip place="top">Hover Me</Tooltip>);
+      const tooltip = mount(
+        <Tooltip place="top" text={TOOLTIP_TEXT}>
+          Hover Me
+        </Tooltip>,
+      );
       expect(tooltip.html()).toMatchSnapshot();
```
allyssonsantos: I think you can use the `shallow` method from enzyme and remove this `.html()`.

alan-oliv: Thanks @allyssonsantos! Lets create a issue so we can review this tests

PR #37, opened 2019-04-01
[BTS-167]: SnackBar Component

rapahaeru:
http://vm-tmb-jirasoft01.servers/browse/BTS-167

### DIFF
--- a/components/SnackBar/SnackBar.jsx
+++ b/components/SnackBar/SnackBar.jsx
@@ -101,9 +101,9 @@ class SnackBar extends React.Component {
   componentWillUnmount() {
     const { body } = document;
     body.removeChild(this.snackBarSection);
+    clearTimeout(this.closeEventtimeOut);
   }
 
-  // this event was implemented to simulate unit tests (a11y)
   handleKeyPress = event => {
     const { onClose } = this.props;
     if (event.key === 'Enter') onClose();
@@ -114,7 +114,7 @@ class SnackBar extends React.Component {
   };
 
   closeOnTime = (secondsToClose, callback) => {
-    setTimeout(() => {
+    this.closeEventtimeOut = setTimeout(() => {
       callback();
     }, secondsToClose * 1000);
   };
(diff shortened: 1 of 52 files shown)


components/SnackBar/SnackBar.jsx
```diff
@@ -0,0 +1,202 @@
...
+
+  componentDidMount() {
+    const { body } = document;
+    body.appendChild(this.snackBarSection);
+
+    this.objectFocus(document.querySelector('.action-button'));
+
+    const { onClose, secondsToClose } = this.props;
+
+    setTimeout(() => {
```
allyssonsantos: I think you must save this `setTimeout` to a class property to "clear" this on componentWillUnmount();

If you open a snackbar and count to 3, click on close icon, open another snackbar and wait, instead of closing in 6 seconds the new one will close after 3 seconds.

components/SnackBar/SnackBar.jsx
```diff
@@ -0,0 +1,202 @@
...
+  handleKeyPress = event => {
+    const { onClose } = this.props;
+    if (event.key === 'Enter') onClose();
+  };
+
+  objectFocus = element => {
+    element.focus();
+  };
+
+  closeOnTime = (secondsToClose, callback) => {
```
allyssonsantos: this setTimeout is equal the one that is on line 98, why did you need this?

components/SnackBar/SnackBar.jsx
```diff
@@ -0,0 +1,202 @@
...
+      text,
+      onClose,
+      closeButtonAriaLabel,
+      actionTrigger,
+      ...rest
+    } = this.props;
+    return ReactDOM.createPortal(
+      <DialogBlock>
+        <Row>
+          <Col xsmall-offset={1} xsmall={10}>
```
allyssonsantos: This values are wrong, the xsmall have only 4 columns, small 6 columns, medium, large and xlarge have 12 columns.

If you test on a small resolution you will see that the snackbar is at the right corner of the viewport.

image

components/SnackBar/SnackBar.jsx
rapahaeru: Oh! I Forgot this line, sorry

PR #38, opened 2019-04-02
[BTS-169]: Pagination component

allyssonsantos:
pagination


rapahaeru:
How should the component behave when it is mobile?
ps: this print was take in 600px of width

Captura de tela de 2019-04-03 17-05-01

allyssonsantos:
> How should the component behave when it is mobile?
> ps: this print was take in 600px of width
> 
> Captura de tela de 2019-04-03 17-05-01

Hi @rapahaeru 
I talked with the UX to see how this component should behave in small resolutions, he will propose a new design for those resolutions

components/Pagination/sub-components/ActionButton.unit.test.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import ActionButton from './ActionButton';
+
+describe('<ActionButton />', () => {
+  describe('Snapshot', () => {
```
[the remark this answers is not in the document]
allyssonsantos: Done!

components/Pagination/sub-components/Page.unit.test.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import Page from './Page';
+
+describe('<Page />', () => {
+  describe('Snapshot', () => {
```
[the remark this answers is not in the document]
allyssonsantos: Done!

components/Pagination/sub-components/Dots.jsx
```diff
@@ -0,0 +1,47 @@
...
+`;
+
+const sharedPseudoElementStyles = `
+  content: '';
+  position: absolute;
+  tpo: 0;
+  ${sharedStyles}
+`;
+
+const Wrapper = styled.div`
```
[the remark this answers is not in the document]
allyssonsantos: The w3c specification shows that you can use block elements inside nav element.
https://www.w3.org/TR/2011/WD-html5-author-20110809/the-nav-element.html#the-nav-element

components/Pagination/Pagination.jsx
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    return pageHref(page);
+  };
+
+  return (
+    <Wrapper aria-label={ariaLabel} {...props}>
+      <ActionButton
+        aria-disabled={activePage === 1}
+        disabled={activePage === 1}
```
ddsilva: It's necessary to add `aria-disabled` prop when there is a `disabled` as well?

allyssonsantos: Removed the `disabled` prop and using `aria-disabled` for the style.

stories/Pagination/Pagination.story.jsx
```diff
@@ -0,0 +1,297 @@
...
+      easier to understand blocks.
+    </Heading>
+    <TabbedView>
+      <Tab title="Usage">
+        <StoryContainer>
+          <Title as="h2">Importing Pagination</Title>
+          <SimpleHighlight>{`import { Pagination } from '@catho/quantum';`}</SimpleHighlight>
+
+          <p>
+            An adaptive Pagination component, if viewport width is below small
```
ggdaltoso:
```suggestion
             The Pagination component will render a mobile-friendly component if the viewport width is below small breakpoint value else it will render a dotted component.
```

:nail_care:

stories/Pagination/Pagination.story.jsx
```diff
@@ -0,0 +1,297 @@
...
+          </p>
+
+          <p>
+            <strong>
+              It&apos;s highly recommended to avoid pagination and use infinite
+              scroll in small devices.
+            </strong>
+          </p>
+
+          <Title as="h2">Examples</Title>
```
ggdaltoso: There is no need for this title, even their description. It is confusing because the description isn't describing the next example (Simple usage) and it is already mentioned [here](https://github.com/catho/quantum/pull/38/files#diff-65ccaff79a6419ef8cc47c36be2ea99bR209), I think.

It may start [here](https://github.com/catho/quantum/pull/38/files#diff-65ccaff79a6419ef8cc47c36be2ea99bR61) btw.

stories/Pagination/Pagination.story.jsx
```diff
@@ -0,0 +1,297 @@
...
+                activePage={5}
+              />
+            </Col>
+          </Row>
+
+          <Title as="h3">Helper text</Title>
+          <Row>
+            <Col medium={5}>
+              <p>
+                When viewport is below <code>small</code> breakpoint, then,
```
ggdaltoso: This info is duplicated at the top of this page. Consider leaving only one.

stories/Pagination/Pagination.story.jsx
```diff
@@ -0,0 +1,297 @@
...
+              </SimpleHighlight>
+            </Col>
+            <Col medium={7}>
+              <Pagination totalPages={10} pageHref={page => `/?page=${page}`} />
+            </Col>
+          </Row>
+
+          <Title as="h3">onClickPage prop</Title>
+          <p>
+            Thorugh <code>onClickPage</code> prop is possible to get the page
```
ggdaltoso:
```suggestion
            Through <code>onClickPage</code> prop is possible to get the page
```

stories/Pagination/Pagination.story.jsx
```diff
@@ -0,0 +1,297 @@
...
+            This prop must be a function and it{' '}
+            <strong>receives the page clicked number</strong>.
+          </p>
+          <p>
+            If you set <code>onClickPage</code> prop the <code>pageHref</code>{' '}
+            prop will be ignored.
+          </p>
+          <p>
+            Here&apos;s an example of a controlled pagination that use this prop
+            to handle the state.
```
ggdaltoso:
```suggestion
            to handle the state:
```

components/Pagination/sub-components/ActionButton.jsx
```diff
@@ -0,0 +1,52 @@
...
+  border: 1px solid transparent;
+  border-radius: 10px;
+  cursor: pointer;
+  font-weight: normal;
+  margin: 0 4px;
+  outline: none;
+  padding: 8px;
+  text-align: center;
+  text-decoration: none;
+  transition: color 0.2s ease-in-out, background-color 0.2s ease-in-out,
```
ggdaltoso: Do we have transition/animation definitions?

components/Pagination/sub-components/ActionButton.jsx
```diff
@@ -0,0 +1,52 @@
...
+    border 0.2s ease-in-out;
+  user-select: none;
+  white-space: nowrap;
+
+  ${({ 'aria-disabled': ariaDisabled }) =>
```
ggdaltoso: Why not `disabled`?

allyssonsantos: It was disabled in the past, but, the `disabled` prop was only for the styling porpoise, using `aria-disabled` we removed disabled prop since this `aria-disabled` needs to be present when the ActionButton is disabled (for accessibility improvements), so, talking with @ddsilva we decided that using only this prop, even for style or accessibility would be better.

components/Pagination/sub-components/ActionButton.jsx
allyssonsantos: No.

stories/Pagination/Pagination.story.jsx
allyssonsantos: Done.
Thanks!

stories/Pagination/Pagination.story.jsx
allyssonsantos: Done.
Thanks!

stories/Pagination/Pagination.story.jsx
allyssonsantos: Done.
Thanks!

stories/Pagination/Pagination.story.jsx
allyssonsantos: Done.
Thanks!

stories/Pagination/Pagination.story.jsx
allyssonsantos: Done.
Thanks!

components/Pagination/sub-components/ActionButton.jsx
ggdaltoso: So, why those transitions?

components/Pagination/sub-components/ActionButton.jsx
ggdaltoso: I'll not apt for disabling an `ActionButton` only using a disabled prop? Must I use `aria-disabled` instead?

allyssonsantos: Right, since `disabled` attribute do not have any behavior or semantic meaning to an anchor element.
`disabled` attribute only works to form elements.

ggdaltoso: Ok

components/Pagination/sub-components/ActionButton.jsx
allyssonsantos: We have transition in several components, like `Button`, `Checkbox`, `Tooltip`, etc.
It's just to be a little bit smooth, what do you think?

Or should we remove the transitions from all components, since we do not have transition/animations definition to none of them?

ggdaltoso: Really don't know. 

We just need to be cautious since we add this by our own. IMO, it could be added to all components but, we always need to keep the UX team in alignment with this.

I don't think we should remove the transitions from all components. Instead, we should avoid adding it to the newers.

allyssonsantos: It was aligned with the UX team, he made his review of how the usability is, and, for him, it's okay.

As next steps I think we need to get this definitions of animations documented by the UX team to get this centralized and make reusable through all components.

components/Pagination/Pagination.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const handleHref = page => {
+    if (!pageHref || onPageClick || page <= 0 || page > totalPages) {
+      return undefined;
+    }
+
+    return pageHref(page);
+  };
+
+  const width = (window && window.innerWidth) || 480;
```
alan-oliv: Can we use BREAKPOINTS.small instead of '480' ? <3

components/Pagination/Pagination.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+Pagination.defaultProps = {
+  ariaLabel: 'pagination',
+  activePage: 1,
+  activePageAriaLabel: 'Current Page',
+  nextButtonText: 'Next',
+  pageAriaLabel: 'Page',
+  pageHref: undefined,
+  prevButtonText: 'Previous',
+  onPageClick: undefined, // eslint-disable-line
```
alan-oliv: I removed this rule disable and nothing happened... is this really necessary?

allyssonsantos: It's not necessary anymore
Done!

components/Pagination/Pagination.jsx
allyssonsantos: Yes, of course!
Done!

PR #39, opened 2019-04-03
[BTS-165] Radio button group component

ddsilva:
radio-button-group


components/RadioGroup/RadioButton.jsx
```diff
@@ -0,0 +1,106 @@
...
+        checked={checked}
+        disabled={disabled}
+        id={_id}
+        onChange={e => onChange({ value, label }, e)}
+        value={value}
+        skin={skin}
+        error={error}
+        {...rest}
+      />
+      <ButtonGroupLabel
```
rapahaeru: Is too much insert a `role="button"` on this label?
Because in this scenario, its plays the role of a button.
Do you agree?

components/RadioGroup/RadioGroup.unit.test.jsx
```diff
@@ -5,6 +5,78 @@ import diff from 'jest-diff';
...
+  describe(`<${RadioItem.displayName} />`, () => {
+    it(`should render three <${RadioItem.displayName} />`, () => {
+      const component = (
+        <RadioGroup name="foo">
+          <RadioItem value="Foo">Foo</RadioItem>
+          <RadioItem value="Bar">Bar</RadioItem>
+          <RadioItem value="Baz">Baz</RadioItem>
+        </RadioGroup>
+      );
+      const wrapper = shallow(component);
```
rapahaeru: [IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=bo_efYhYU2A)

components/shared/HiddenRadio.jsx
```diff
@@ -0,0 +1,38 @@
+import styled, { css } from 'styled-components';
+import ButtonGroupLabel from './ButtonGroupLabel';
+import { ERROR_HOVER_STYLE } from './inputStyle';
+import { FOCUS_STYLE, HOVER_STYLE, ACTIVE_STYLE } from '../Button/buttonStyle';
```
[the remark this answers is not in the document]
ddsilva: I think I got your point.
What if we extract button style to the shared folder as we did with the input style?

components/Button/buttonStyle.js
```diff
@@ -0,0 +1,141 @@
...
+    large: '48px',
+    xlarge: '56px',
+  };
+
+  return `height: ${heights[size] || heights.medium};`;
+};
+
+const _interactiveStyle = (props, skinProp) => {
+  const {
+    [skinProp]: { shadow, background, border, color },
```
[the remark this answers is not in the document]
ddsilva: Here it's applied a [computed property name](https://medium.com/front-end-weekly/javascript-object-creation-356e504173a8), to get `hovered`, `focused` and `selected` styles, with comes from [this file](https://github.com/catho/quantum/blob/master/components/Button/skins.js#L1), according button skin (`primary`, `secondary`, etc).

components/RadioGroup/RadioButton.jsx
ddsilva: I don't know if I agree, because we only apply button skin here. The element still a radio input. WDYT?

components/shared/HiddenRadio.jsx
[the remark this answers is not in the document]
rapahaeru: Sounds good. But what name we do to that file?

components/RadioGroup/RadioButton.jsx
rapahaeru: i dont know to say...
That`s a issue.
Anyway, lets keep like this.

components/shared/HiddenRadio.jsx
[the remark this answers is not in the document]
ddsilva: I think we can implement this on themification task.

components/RadioGroup/RadioButton.jsx
```diff
@@ -0,0 +1,106 @@
...
+  onChange,
+  value,
+  checked,
+  icon,
+  id,
+  inline,
+  ...rest
+}) => {
+  const skin = checked ? 'primary' : 'secondary';
+  const _id = id || uniqId('radio-button-');
```
alan-oliv: Still about generators, this line would look somethings like this:

```suggestion
const _id = id || ID_GENERATOR.next().value;
```

components/RadioGroup/RadioButton.jsx
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import ButtonGroupLabel from '../shared/ButtonGroupLabel';
+import HiddenRadio from '../shared/HiddenRadio';
+import Icon from '../Icon';
+import uniqId from '../shared/uniqId';
+
```
alan-oliv: What you think about using a generator function in uniqId module?
We could add these 2 lines here:

```suggestion
const DEFAULT_ID_PREFIX = 'radio-button-';
const ID_GENERATOR = uniqId(DEFAULT_ID_PREFIX);
```

components/RadioGroup/RadioButton.jsx
ddsilva: Done

components/RadioGroup/RadioButton.jsx
ddsilva: Done

PR #50, opened 2019-04-18
[BTS-209] Autocomplete

ggdaltoso:
I used to say Typeahead but, really, it need to be called "Autocomplete".

It is just a prop from Dropdown, as it is beautifully implemented in Downshift.

Autocomplete


components/Dropdown/Dropdown.jsx
```diff
@@ -141,6 +149,9 @@ const Dropdown = ({
     return changes;
   };
 
+  const inputFilter = value =>
```
alan-oliv: <3

allyssonsantos:
:heart:

PR #60, opened 2019-05-02
[FIX] Sanitize title in using HTML attributes in <TabbedView />

tcelestino:
When there are title with accents and space in `<Tab />`, they're set in attributes in the elements.

Screenshot from 2019-05-02 15-43-13

This pull request removes spaces and accents. It will replace space with "-" too.

### DIFF
--- a/components/TabbedView/TabbedView.jsx
+++ b/components/TabbedView/TabbedView.jsx
@@ -101,6 +101,13 @@ class TabbedView extends React.Component {
     this.setState({ activeTab: tab });
   };
 
+  sanitize = str =>
+    str
+      .normalize('NFD')
+      .replace(/[\u0300-\u036f]/g, '')
+      .replace(' ', '-')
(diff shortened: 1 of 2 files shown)


components/TabbedView/TabbedView.jsx
```diff
@@ -113,8 +120,8 @@ class TabbedView extends React.Component {
               key={title}
               onClick={() => this.onTabClick(title)}
               skin={skin}
-              id={`${title}-tab`}
-              aria-controls={`${title}-panel`}
+              id={`${this.sanitize(title)}-tab`}
```
ddsilva: What do you think of using this [module](https://github.com/catho/quantum/blob/master/components/shared/uniqId.js) to generate the tab id?
[Here](https://github.com/catho/quantum/blob/master/components/Input/Input.jsx#L128) is an example of how to use it.

PR #73, opened 2019-05-24
[BTS-345] Apply theme props to <Tooltip />

allyssonsantos:


components/Tooltip/Tooltip.jsx
```diff
@@ -86,11 +104,20 @@ Tooltip.propTypes = {
     PropTypes.arrayOf(PropTypes.node),
     PropTypes.node,
   ]).isRequired,
+  theme: PropTypes.shape({
+    spacing: PropTypes.object,
+    colors: PropTypes.object,
```
ggdaltoso:
```suggestion
    colors: PropTypes.object,
    baseFontSize: PropTypes.number,
```

allyssonsantos: done

PR #81, opened 2019-05-29
[BTS-275] Applying theme props to Checkbox

ddsilva:


components/Checkbox/Checkbox.jsx
```diff
@@ -38,69 +46,110 @@ const HiddenCheckbox = styled(HiddenInput).attrs({
...
+      }
+
+      :checked + ${CheckIcon} {
+        background-color: ${neutral[500]};
+      }
+    `}
 
     :hover + ${CheckIcon} {
-      box-shadow: none;
+      ${shadow(0)}
```
ggdaltoso: Isn't 0 the default elevation value? If yes, is it necessary here?

components/Checkbox/Checkbox.jsx
```diff
@@ -18,9 +18,17 @@ const CheckboxWrapper = styled.div`
...
+    },
+  }) => neutral[700]};
   font-size: 16px;
-  margin: 0 0 0 10px;
+  margin: ${({
+    theme: {
+      spacing: { xsmall },
+    },
+  }) => `0 0 0 ${xsmall}px`};
 `;
```
allyssonsantos: What do you think to put the color and margin callbacks together?

```suggestion
const CheckboxLabel = styled(Label)`
  ${({ theme: { colors: { neutral }, spacing: { xsmall } } }) => `
    color: ${neutral[700]};
    margin: 0 0 0 ${xsmall}px;
  `}
  font-size: 16px;
`;
```

components/Checkbox/Checkbox.jsx
```diff
@@ -38,69 +46,110 @@ const HiddenCheckbox = styled(HiddenInput).attrs({
...
     justify-content: center;
-    margin-top: 2px;
+    margin-top: ${({
+      theme: {
+        spacing: { xxxsmall },
+      },
+    }) => `${xxxsmall}px`};
     transition: all 0.2s ease-in-out;
     width: ${CHECKBOX_SIZE};
+
```
allyssonsantos: And here too.

components/Checkbox/Checkbox.jsx
```diff
@@ -38,69 +46,110 @@ const HiddenCheckbox = styled(HiddenInput).attrs({
 
   + ${CheckIcon} {
     align-items: center;
-    background-color: ${Colors.WHITE};
     border-radius: 2px;
-    border: 2px solid ${Colors.BLACK['400']};
     box-sizing: border-box;
     color: transparent;
     display: flex;
     font-size: 16px;
```
alan-oliv: There's a reason not to use baseFontSize from theme here?

components/Checkbox/Checkbox.jsx
alan-oliv: And baseFontSize too (if there's not a reason)

PR #85, opened 2019-06-03
[BTS-274] button

ggdaltoso:


components/Button/Button.jsx
```diff
@@ -1,91 +1,329 @@
...
+
+  ${({ size, theme: { baseFontSize } }) => {
+    const sizes = {
+      xsmall: `${baseFontSize * 0.75}px`,
+      small: `${baseFontSize * 0.75}px`,
+      medium: `${baseFontSize}px`,
+      large: `${baseFontSize * 1.25}px`,
+      xlarge: `${baseFontSize * 1.5}px`,
+    };
+    return `font-size: ${sizes[size] || sizes.medium};`;
```
alan-oliv: `|| sizes.medium` - Why is this needed (since defaultProps has size: 'medium') ?
(Same doubt in `height` and `padding` props)

components/Button/Button.jsx
```diff
@@ -1,91 +1,329 @@
...
+  }}
+
+  ${({
+    size,
+    theme: {
+      spacing: { large, xlarge, xxlarge, xxxlarge },
+    },
+  }) => {
+    const heights = {
```
alan-oliv: Is this property really needed? Can we solve with padding only?
If can't, IMHO I'd avoid using spacing values for component height

components/Button/Button.jsx
```diff
@@ -1,91 +1,329 @@
...
+      background-color: ${bgColor};
+      color: ${textColor};
+
+      border: 1.5px solid ${disabled ? neutral500 : mainColor500};
+
+      ${shadow(2, neutral500)({ theme })}
+
+      :hover {
+        ${
+          !disabled
```
alan-oliv: What do you guys think of using an logical and (&&) here instead of a ternary, since has no else condition?
(in `focus` and `active` too)

components/Button/Button.jsx
```diff
@@ -1,91 +1,329 @@
...
+      ${({
+        size,
+        theme: {
+          spacing: { medium, large },
+        },
+      }) => {
+        const sizes = {
+          xsmall: `${medium}px`,
+          small: `${medium}px`,
+          normal: `${large}px`,
```
alan-oliv: I guess we can update this object and rename `normal` to `medium` to keep with our nomenclature convention

components/Button/Button.jsx
```diff
@@ -1,91 +1,329 @@
...
+  ${props =>
+    props.center &&
+    `
+    margin-left: auto;
+    margin-right: auto;
+  `}
 
   ${ButtonIcon} {
-    ${iconSize}
+    margin-right: 5px;
```
ddsilva: Is `spacing.xxsmall` (4px) not electable in place of this hard coded value?

components/Button/Button.jsx
ggdaltoso: There is no need for this. We get the point. :+1:

components/Button/Button.jsx
ggdaltoso: _Is this property really needed?_
Yes. It is needed.

_Can we solve with padding only?_
Yes, we can.

_If can't, IMHO I'd avoid using spacing values for component height_
We follow material design (MD) guidelines as a reference for this. In this case, the MD uses height to define the vertical size of the component.

components/Button/Button.jsx
ggdaltoso: Yes, it is :clap: 

@allyssonsantos

components/Button/Button.jsx
ggdaltoso: Using `&&` will return a boolean condition what will make a string template to be like this:
 ```
   :hover { false; }
```
We use a ternary expression to prevent that situation.

components/Button/Button.jsx
ggdaltoso: :+1:

components/Button/Button.jsx
alan-oliv: If we can solve with padding, so it is not needed (respectin')
I'd remove this part, but i'll leave to you guys to decide or vote <3

ddsilva: Although Material Design specifies that this is solved with height, it's a common sense among us (including me) that padding it's better for this case. So, is there any problem to implements in this way?

PR #97, opened 2019-06-21
fix(window-server-side-rendering): Change validation the window in Pa…

MichelAraujo:
In nodejs (server side rendering) don't exist variable window and throw an error because this (not work), then is need changed the validation to work in server side rendering.

### DIFF
--- a/components/Pagination/Pagination.jsx
+++ b/components/Pagination/Pagination.jsx
@@ -47,7 +47,10 @@ const Pagination = ({
     return pageHref(page);
   };
 
-  const width = (window && window.innerWidth) || BREAKPOINTS.small.width;
+  let { width } = BREAKPOINTS.small;
+  if (typeof window !== 'undefined') {
+    width = window.innerWidth;
+  }
 
   const Component = width > BREAKPOINTS.small.width ? Desktop : Mobile;
 


ddsilva:
Nice, @MichelAraujo!

I'm wondering if it's possible to write a unit test for this case, maybe creating a mocked window object and making a snapshot test with it.

What do you think?

MichelAraujo:
> Nice, @MichelAraujo!
> 
> I'm wondering if it's possible to write a unit test for this case, maybe creating a mocked window object and making a snapshot test with it.
> 
> What do you think?

Sure, I'II make a unit test for this.

MichelAraujo:
How I make a mock of the windows variable with value undefined ?

MichelAraujo:
@rapahaeru can validate this PR

renatogalvones:
@MichelAraujo  Here's an example on how to mock the window object.

```
/* eslint-disable no-undef */
const mockWindow = {
  btoa() {},
  matchMedia() {
    return { matches: {} };
  },
  addEventListener() {},
  removeEventListener() {},
  setTimeout() {},
  requestAnimationFrame() {},
  cancelAnimationFrame() {},
  scrollTo() {},
  scrollBy() {},
  LO: {
    page() {},
  },
  hbspt: {
    forms: {
      create() {},
    },
  },
  jQuery: {},
  // Add more fields from window as necessary
};

export { mockWindow };

const defaultWindow = typeof window !== 'undefined' ? window : mockWindow;

export default defaultWindow;
```

PR #103, opened 2019-08-14
Fix/button height

pedrohmorais:
## WHAT THIS PR DOES
Fixes the button component height when it have more than two lines.

## HOW TO REPRODUCE
Import the button component and force it to have two or more lines.


components/Button/Button.jsx
```diff
@@ -59,10 +59,10 @@ const StyledButton = styled.button`
     },
   }) => {
     const paddings = {
```
alisson-suzigan: The papding values should not be multiples of two?

ggdaltoso: And it should be getting from the theme

pedrohmorais: A solution has been proposed in the last commit.

PR #104, opened 2019-08-15
fix: fixes input component events onClean and onChange

pedrohmorais:
## WHAT THIS PR DOES
Fixes the `onClean` event on `Input` component and add the event `onChange` on it.

### DIFF
--- a/components/Input/Input.jsx
+++ b/components/Input/Input.jsx
@@ -89,11 +89,28 @@ class Input extends React.Component {
   constructor(props) {
     super(props);
 
-    const { type, id } = props;
+    const { type, id, value } = props;
 
-    this.state = { type };
+    this.state = {
+      type,
+      isDirty: !!value,
+      currentValue: value,
+    };
 
     this._id = id || ID_GENERATOR.next().value;
+
+    this.onChangeInput = this.onChangeInput.bind(this);
+  }
+
+  onChangeInput(ev) {
+    const { onChange } = this.props;
+    const vl = ev.currentTarget.value;
+    this.setState({
+      currentValue: vl,
+      isDirty: !!vl[0],
+    });
+
+    onChange();
   }
 
   _changeType = type => {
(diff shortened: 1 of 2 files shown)


components/Input/Input.jsx
```diff
@@ -89,11 +89,28 @@ class Input extends React.Component {
   constructor(props) {
     super(props);
 
-    const { type, id } = props;
+    const { type, id, value } = props;
 
-    this.state = { type };
+    this.state = {
+      type,
+      isDirty: !!value,
```
renatogalvones: `hasValue` would be easier to understand

components/Input/Input.jsx
```diff
@@ -89,11 +89,28 @@ class Input extends React.Component {
...
+    };
 
     this._id = id || ID_GENERATOR.next().value;
+
+    this.onChangeInput = this.onChangeInput.bind(this);
+  }
+
+  onChangeInput(ev) {
+    const { onChange } = this.props;
+    const vl = ev.currentTarget.value;
```
renatogalvones: `inputValue` instead `vl` would be easier to read

components/Input/Input.jsx
renatogalvones: Is this state really necessary? Maybe you can validate this just using `currentValue`

components/Input/Input.jsx
```diff
@@ -89,11 +89,26 @@ class Input extends React.Component {
...
 
-    this.state = { type };
+    this.state = {
+      type,
+      currentValue: value,
+    };
 
     this._id = id || ID_GENERATOR.next().value;
+
+    this.onChangeInput = this.onChangeInput.bind(this);
```
ggdaltoso: Why not use [public class fields](https://github.com/tc39/proposal-class-fields) instead of binding functions?

This project adopts this pattern to make the code clean and there is no reference of binding events/methods/functions in any source code.

pedrohmorais: @ggdaltoso fixed in last commit.

PR #105, opened 2019-08-17
feat: create 2 new colors in pallete neutral and change in some compo…

pedrohmorais:
👍    @rapahaeru 
## What this PR does
Creates two new colors in pallete and fix the white color in some components.


components/shared/theme.js
```diff
@@ -22,46 +22,58 @@ const breakpoints = {
```
allyssonsantos: Why do you have white and black in all palette colors?
Does not sound weird to have a white danger color, or a secondary black color?

And, I see that you are not using the commitizen `(yarn commitizen)` to manage your commits, It would be nice to use it, the semantic-release read them and put in the release notes based on the conventional commit pattern.

https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines

pedrohmorais: I inserted this colors in all palette just to prevent issues if a different theme try to use the 0 or 1000 color.

@allyssonsantos do you think it would be better to do otherwise and remove the 0 and 1000 from remainings palletes ?

And thanks to warning me to use the `yarn commitizen`.

rapahaeru:
I believe its popover must be updated in theme.js, in his node.
what do you think about update this way:

```
popover: {
    skins: {
      neutral: {
        background: colors.neutral[0],
```

on Zeplin, the neutral color for that component, is white and still showing gray.

rapahaeru:
In Radio group, the background still gray.

Captura de tela de 2019-09-04 15-54-32

Zeplin:
Captura de tela de 2019-09-04 15-55-23

I believe that problem resolves on Radio.jsx, probably the neutral color still `100` instead `0`

rapahaeru:
Same here.
Captura de tela de 2019-09-04 16-03-58

Ps: This component, affects directly the buttonGroup component. 
I believe its resolves another problem on the buttonGroup component:
Captura de tela de 2019-09-04 16-05-46

rapahaeru:
For documentation, i think its necessary to update the color palette section, on the **Foundation > Color palette**

https://catho.github.io/quantum/?path=/story/foundation--color-palette

That colors in there are deprecated.

in that, could be good if we update it together, what do you think?

pedrohmorais:
@rapahaeru all your recomendations are in commit ffd2e1430f59dfe32711537317e0839ab9107e29

stories/Colors/Colors.jsx
```diff
@@ -198,6 +198,11 @@ export default () => (
     <Title as="h3" style={{ paddingTop: 0 }}>
       Black
     </Title>
+    <p>
```
rapahaeru: The new colors update on documentation is fine, but we need do update all the other colors too.
Notice that colors and his names is all deprecated. we use six colors level now (except the neutral).
We can do it together and show to Arthur, what do you think?

rapahaeru:
In Button.jsx, the text still gray.
i think is only update the line 121 to use the color `0`

image

PR #108, opened 2019-08-22
fix(pagination): implements tabIndex in pagination components to fix …

pedrohmorais:
## What this PR does
Insert tabIndex to inner buttons inside `pagination` component to fix tab navigation issues.

## How to test
Run storybook and open the pagination page, after press tab button an try to navigate into pagination.


components/Pagination/sub-components/Desktop.jsx
```diff
@@ -35,6 +36,7 @@ const Desktop = ({
 
       return (
         <PageButton
+          tabIndex={0}
```
MichelAraujo: This value always be zero? Is your order already defined relative to the other elements?

pedrohmorais: @MichelAraujo now the childrens `tabIndex` prop is inherited from the parent component.
See f9fdfe75b894115005a9f410d3b72265fa169f04.

PR #122, opened 2019-09-04
fix(jest): fix alert tests

pedrohmorais:
✋   @rapahaeru 
## This PR fixes the warnings that appear during jest unit tests.


components/Button/Button.jsx
```diff
@@ -206,9 +206,9 @@ const StyledButton = styled.button`
   }}
 `;
 
-const Button = ({ children, icon, size, $as, ...rest }) => (
-  <StyledButton as={$as} {...rest} size={size}>
-    {icon && <ButtonIcon size={size} name={icon} {...rest} />}
+const Button = ({ children, icon, size, $as, skin, theme, ...rest }) => (
```
rapahaeru: Testing here, i discover that `skin` is not necessary do destruct, only `theme`.
Do you have any reason to apply the `skin` here?

PR #125, opened 2019-09-05
feat(circularLoader): Creates CircularLoader component

pedrohmorais:
👍  @rapahaeru 
## Creates the `CircularLoader` component


rapahaeru:
I think the component story must be updated with de type of new props.

image

PR #127, opened 2019-09-06
Feat/progress bar

pedrohmorais:
:eyes:  @rapahaeru 
## What this PR does
Creates `ProgressBar` component.


rapahaeru:
I think the font default for this component should be 16px instead 12px.
Or i got this wrong?

I believe the subtitle should be a link with some action (switch the exhibition method).
Is not it?
image

components/ProgressBar/ProgressBar.jsx
```diff
@@ -0,0 +1,150 @@
...
+          skins: {
+            [skin]: { background, titleColor, textColor },
+          },
+        },
+      },
+    },
+  } = props;
+
+  return (
+    <Wrapper baseFontSize={baseFontSize} textColor={textColor}>
```
rapahaeru: This `baseFontSize` prop is necessary?
Cause ai think there will be no different font sizes in this component...
What do you think?

components/ProgressBar/ProgressBar.jsx
```diff
@@ -0,0 +1,150 @@
...
+    </Wrapper>
+  );
+};
+
+ProgressBar.defaultProps = {
+  title: null,
+  subTitle: null,
+  width: '250px',
+  height: '8px',
+  progressText: 0,
```
rapahaeru: I have a doubt.
Why the props progressText and progressPercent cant be a unique prop?

components/ProgressBar/ProgressBar.jsx
```diff
@@ -0,0 +1,150 @@
...
+  progressPercent: 0,
+  skin: 'primary',
+  theme: {
+    baseFontSize: 12,
+    components: {
+      progressBar: components.progressBar,
+    },
+  },
+  label: '%',
+  withLabel: true,
```
rapahaeru: I could'nt simulate this prop and looking for zeplin, i think there is no option without label, but i need to confirm

image

stories/Loaders/Loaders.story.jsx
```diff
@@ -16,10 +16,10 @@ const descriptions = {
   circularLoader: `CircularLoaders are used to give the user the impression of loading while a component is not ready to be displayed.`,
 };
 
-const progressBarExamplesTab = (
+const renderExampleTab = examples => (
```
rapahaeru: how about change the examples var to `componentExample` to make it clearer ?
WDYT?

.storybook/regression-test/stories.js
```diff
@@ -19,5 +19,6 @@ export default [
   'Tag',
   'Pagination',
   'Grid',
+  'ProgressBar',
   'CircularLoader',
```
renatogalvones: Why this PR(progress-bar) is changing snapshots of `circularLoader` ?
(I'm mentioning it here, but look on the `png files`

stories/Loaders/Examples/ProgressBar.jsx
```diff
@@ -0,0 +1,39 @@
...
+import ProgressBar from '../../../components/ProgressBar';
+
+export default [
+  {
+    code: '<ProgressBar />',
+    component: <ProgressBar />,
+  },
+  {
+    code: '<ProgressBar progressPercent={30} />',
+    component: <ProgressBar progressPercent={30} />,
```
renatogalvones: On the unit test you should add a test passing props like that.

pedrohmorais: Probably because the tests are outdated with the master branch.

.storybook/regression-test/stories.js
pedrohmorais: Probably because the tests are outdated with the master branch.

PR #133, opened 2019-09-13
build(RegressionTest): Adds test to prepush

renatogalvones:
## Steps to reproduce

1. `git checkout -b test/delete-me-after-this-test`
2. Change a component that will trigger changes on regression test
3. `git push origin test/delete-me-after-this-test`
4. The `pre-push` script must fail
5. Delete your changes
6. Try to push again
7. It should work


package.json
```diff
@@ -12,7 +12,7 @@
     "test:regression:approve": "yarn loki approve --port 9006",
     "test:regression:update": "yarn loki update --port 9006",
     "test:regression:update:storie": "yarn loki update --port 9006 --storiesFilter $STORIE",
-    "test:regression:storybook": "start-storybook -s static -c .storybook/regression-test --port 9006",
+    "test:regression:storybook": "fuser -k 9006/tcp; start-storybook -s static -c .storybook/regression-test --port 9006",
```
rapahaeru: Do you know why this alert prints here?
image

Issue #134, opened 2019-09-17
onChange of <Input> no longer pass event object


Agezao:
Hi,

We use Quantum in some projects here @catho-lab and recently I've upgraded our quantum package to the latest version and started to experience some strange behaviors.

Troubleshooting the problem it turned out that the culprit was one change of behavior at the `onChange` event at <Input /> component. The `onChange` event no longer sends back the event object, nor the changes.

`onChange` call at my previous version:
https://github.com/catho/quantum/blob/9a6e8b85b59ca9255a89f082d40116b333cd0377/components/Input/Input.jsx#L152

`onChange` call at master:
https://github.com/catho/quantum/blob/debce31e7bc002342a7efc59f13c269f663610b0/components/Input/Input.jsx#L109

Is this intentional?

Right now I'm rolling back to the last version that had this event return pattern, but it's definitely not ideal since I'll be losing all the bug-fixes and improvements of the recent versions.

Thanks.

Agezao:
If this is indeed an expected behavior, I've added #135 to address it.

PR #135, opened 2019-09-17
Adding event payload and changed value back onChange event at <Input />

Agezao:
As pointed in #134, this PR intends to add back the behavior of the onChange event of the <Input /> component to return the event object and new payload to the parent callback.

If this is an intentional behavior, please ignore this PR.

### DIFF
--- a/components/Input/Input.jsx
+++ b/components/Input/Input.jsx
@@ -106,7 +106,7 @@ class Input extends React.Component {
       currentValue: inputValue,
     });
 
-    onChange();
+    onChange(ev, { inputValue });
   };
 
   _changeType = type => {


components/Input/Input.jsx
```diff
@@ -106,7 +106,7 @@ class Input extends React.Component {
       currentValue: inputValue,
     });
 
-    onChange();
+    onChange(ev, { inputValue });
```
pedrohmorais: You really need the second parameter ? Because he is already inside the `ev` variable.

pedrohmorais:
@Agezao we get your point and created a new PR with updates that can help you in your issue.
We are closing this PR. Please se the new in https://github.com/catho/quantum/pull/136

components/Input/Input.jsx
Agezao: No, I don't. 
I just keep it for consistency with the previous version that had it, since there might be other projects depending on this second param. But, IMO, he can be disposed.

If you guys agree with the absence of the second parameter I can remove it and update the PR.

PR #138, opened 2019-09-23
feat(snackbar): adds examples to snackbar stories

pedrohmorais:
# What this PR does
Adds examples to SnackBar stories.


rapahaeru:
if you move your action buttons snackbar to the example tab?
There is already an example there..
What do you think?
Just move...

PR #142, opened 2019-09-24
fix(progressbar): fix progressbar issue

pedrohmorais:
## Description
Fix a issue that breaks quantum when it is builded.

## Setup
To view this error, just install and import quantum in a application. Or try to use a builded version.

## Review guide
      - `yarn test:coverage`
- [x] Regression \
      - first start the storybook for regression tests: ` yarn test:regression:storybook`; \
- [] Code review

### Browsers review
- [x] Layout review
  - [x] Edge
  - [x] IE 11
  - [x] Firefox
  - [x] Chrome
  - [x] Safari
  - [x] Mobile
- [ ] Ux review validation


components/shared/hexToRgb.js
```diff
@@ -0,0 +1,11 @@
+const hexToRgb = hex => {
```
rapahaeru: Testing the implementation, i noticed that method is not used in any place.
What do you think to just remove it?

components/shared/index.js
```diff
@@ -8,6 +8,7 @@ import BREAKPOINTS from './breakpoints';
 import uniqId from './uniqId';
 import * as INPUT_STYLE from './inputStyle';
 import hexToRgba from './hexToRgba';
+import hexToRgb from './hexToRgb';
```
rapahaeru: we don't need this anymore, if i am correct.

renatogalvones: you're correct! :wink:

components/shared/hexToRgb.js
pedrohmorais: Good idea!

components/shared/index.js
pedrohmorais: Removed

PR #162, opened 2019-10-23
fix/radio-button: update component with the style guide

rapahaeru:
## Description
https://jirasoftware.catho.com.br/browse/QTM-67

## Setup

## Review guide
- [ ] Regression \
- [ ] Code review

### Browsers review
- [ ] Layout review
  - [ ] Edge
  - [ ] IE 11
  - [ ] Firefox
  - [ ] Chrome
  - [ ] Safari
  - [ ] Mobile
- [ ] Ux review validation


components/RadioGroup/Radio.jsx
```diff
@@ -4,37 +4,39 @@ import styled from 'styled-components';
 import { Label, HiddenInput, shadow } from '../shared';
 import { colors, spacing } from '../shared/theme';
 
+const radioSize = '24px';
```
renatogalvones: What do you think about using `spacing.large` here?

rapahaeru: i guess its not a good idea @renatogalvones , cause the spacing is based in gutter point. The gutter can be changed and if this is done, will be changed the size of components too. I think its not right.
And the spacings are used only for paddings and margins.
Do you agree?

PR #173, opened 2019-12-03
Fix/circular loader

pedrohmorais:
## Description
Fix `CircularLoader` on IE and Firefox.

## Setup

## Review guide
      - `yarn test:coverage`
- [ ] Regression \
- [ ] Code review

### Browsers review
- [ ] Layout review
  - [ ] Edge
  - [ ] IE 11
  - [ ] Firefox
  - [ ] Chrome
  - [ ] Safari
  - [ ] Mobile
- [ ] Ux review validation

### DIFF
--- a/components/CircularLoader/CircularLoader.jsx
+++ b/components/CircularLoader/CircularLoader.jsx
@@ -101,13 +101,7 @@ const CircularLoader = props => {
   return (
     <Wrapper size={size} gutter={gutter} role="progressbar">
       <Content color={color} viewBox="22 22 44 44">
-        <Circle
-          STROKE-WIDTH={strokeWidth}
-          cx={cx}
-          cy={cy}
-          r={r}
-          color={color}
-        />
+        <Circle strokeWidth={strokeWidth} cx={cx} cy={cy} r={r} color={color} />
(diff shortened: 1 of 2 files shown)


components/CircularLoader/CircularLoader.jsx
```diff
@@ -87,10 +96,18 @@ const CircularLoader = props => {
...
   } = props;
 
+  const { cx, cy, r, strokeWidth } = circleProps;
+
   return (
     <Wrapper size={size} gutter={gutter} role="progressbar">
       <Content color={color} viewBox="22 22 44 44">
-        <Circle color={color} />
+        <Circle
+          STROKE-WIDTH={strokeWidth}
```
alisson-suzigan:
```suggestion
          stroke-width={strokeWidth}
```

pedrohmorais: This prop is following the material-ui, i have to check if it will not break IE.

alisson-suzigan: Ok, the ideia is just to move the property name to lower case. Is it a problem?

PR #183, opened 2020-02-07
[DOCS] Updated README to reference the types package

MutterPedro:
## Description
Updated README to reference the definition of the types


MutterPedro:
Não tenho permissão pra fazer o merge :disappointed:

PR #186, opened 2020-02-12
feat(Dropdown): Add property to ignore special chars

MichelAraujo:
## Description
The current version of the component Dropdown when used how autocomplete don't return list if input is not identical, exemple with a itens list how [ 'são paulo',  'belém'] just return if input is 'são paulo' or 'belém'.
With this parameter "ignoreSpecialChars" we can input 'sao paulo' and return list the suggest corrctly.

Changes list:
* Add new parameter in Dropdown component
* Create a function to remove special chars inside the Dropdown component
* Add unit test for Dropdown autocomplete with and without new prop
* Add new parameter in doc storybook 

## Review guide
      - `yarn test:coverage`
- [ ] Regression \
- [ ] Code review

### Browsers review
- [ ] Layout review
  - [x] Edge
  - [x] IE 11
  - [ ] Firefox
  - [x] Chrome
  - [ ] Safari
  - [x] Mobile
- [ ] Ux review validation

### DIFF
--- a/components/Dropdown/Dropdown.jsx
+++ b/components/Dropdown/Dropdown.jsx
@@ -302,6 +302,7 @@ const Dropdown = ({
   autocomplete,
   theme,
   id,
+  ignoreSpecialChars,
   ...rest
 }) => {
   const _buttonLabel = selectedItem ? _getLabel(selectedItem) : placeholder;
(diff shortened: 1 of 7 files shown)


components/Dropdown/Dropdown.jsx
```diff
@@ -325,14 +326,31 @@ const Dropdown = ({
 
   const [_id] = useState(id || ID_GENERATOR.next().value);
 
-  const inputFilter = value =>
-    items.filter(
-      item =>
-        _getValue(item)
-          .toLowerCase()
-          .indexOf(value.toLowerCase()) > -1,
+  const normalizeChars = value =>
```
renatogalvones: Not a block, but would be nice to have this function isolated so we can reuse it on other components.

Suggestion: `Components/shared/` folder. 

Just like `hexToRgba`:

https://github.com/catho/quantum/blob/master/components/shared/hexToRgba.js

MichelAraujo: great idea!
Done!

renatogalvones: :tada:

PR #196, opened 2020-02-21
AUTH-165 Added SocialButton component

MutterPedro:
## Description
Added SocialButton component and its tests.

## Setup

## Review guide
      - `yarn test:coverage`
- [ ] Regression \
- [ ] Code review

### Browsers review
- [ ] Layout review
  - [ ] Edge
  - [ ] IE 11
  - [ ] Firefox
  - [ ] Chrome
  - [ ] Safari
  - [ ] Mobile
- [ ] Ux review validation


stories/Button/Button.story.jsx
```diff
@@ -105,20 +106,56 @@ const exampleTab = (
...
+            <SocialButton provider="facebook" title="facebook-button" />
+          }
+          code={"<SocialButton provider='facebook' title='facebook-button'/>"}
+        />
+      </Container>
+    </StoryContainer>
+  </Tab>
+);
+
 const description = `Buttons express what action will occur when the user clicks
```
rapahaeru: how about we update this description to this specific buttons?

like `this buttons are used to login access with its respective social media.`

components/SocialButton/SocialButton.jsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import Button from '../Button';
+import FacebookIcon from './sub-components/FacebookIcon';
+import GoogleIcon from './sub-components/GoogleIcon';
+import { theme as defaultTheme } from '../shared';
+
+const GoogleButton = styled(Button)`
```
rapahaeru: Suggestion: to reuse and center code, how about this:

```
const BUTTON_SIZE = '40px';
const BUTTON_PADDING = defaultTheme.spacing.xsmall;

const BaseButton = styled(Button)`
  width: ${BUTTON_SIZE};
  height: ${BUTTON_SIZE};
  border-radius: 9px;
  cursor: pointer;
  overflow: hidden;
`;

const GoogleButton = styled(BaseButton)`
  outline: none;
  padding: ${BUTTON_PADDING - 1}px;
`;

const FacebookButton = styled(BaseButton)`
  border: none;
  background-color: #3c5193;
  padding: ${BUTTON_PADDING}px;
`;
```

rapahaeru:
Firefox: 
image

components/SocialButton/SocialButton.jsx
```diff
@@ -0,0 +1,59 @@
...
+    case 'facebook':
+      return (
+        <FacebookButton {...props}>
+          <FacebookIcon title={title} />
+        </FacebookButton>
+      );
+    case 'google':
+      return (
+        <GoogleButton {...props} stroked skin="neutral">
+          <GoogleIcon title={title} size={24} />
```
rapahaeru: this props should be a string instead a number.

components/SocialButton/sub-components/GoogleIcon.unit.test.jsx
```diff
@@ -0,0 +1,17 @@
...
+describe('GoogleIcon component', () => {
+  it('Should match the snapshot', () => {
+    expect(
+      renderer.create(<GoogleIcon title="test" />).toJSON(),
+    ).toMatchSnapshot();
+  });
+
+  it('when size prop is set to 120', () => {
+    expect(
+      renderer.create(<GoogleIcon title="test" size={120} />).toJSON(),
```
rapahaeru: this props should be a string instead a number.

components/SocialButton/sub-components/FacebookIcon.unit.test.jsx
```diff
@@ -0,0 +1,17 @@
...
+describe('FacebookIcon component', () => {
+  it('Should match the snapshot', () => {
+    expect(
+      renderer.create(<FacebookIcon title="test" />).toJSON(),
+    ).toMatchSnapshot();
+  });
+
+  it('when size prop is set to 120', () => {
+    expect(
+      renderer.create(<FacebookIcon title="test" size={120} />).toJSON(),
```
rapahaeru: this props should be a string instead a number.

components/SocialButton/sub-components/FacebookIcon.jsx
```diff
@@ -0,0 +1,54 @@
...
+    <StyledSvg
+      role="img"
+      aria-labelledby="facebook-title"
+      xmlns="http://www.w3.org/2000/svg"
+      width={size}
+      height={size}
+    >
+      <title id="facebook-title">{title}</title>
+      <g>
+        <title>background</title>
```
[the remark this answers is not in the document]
MutterPedro: Do you prefer to keep a static one?

components/SocialButton/sub-components/FacebookIcon.jsx
```diff
@@ -0,0 +1,54 @@
...
+          fill="none"
+          id="canvas_background"
+          height={SVGOriginalHeight}
+          width={SVGOriginalWidth}
+          y="-1"
+          x="-1"
+        />
+      </g>
+      <g>
+        <title>Layer 1</title>
```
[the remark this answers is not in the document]
MutterPedro: This came with the SVG, but I think that could be removed without problem

stories/Button/Button.story.jsx
MutterPedro: Will they be used only for login?

components/SocialButton/SocialButton.jsx
MutterPedro: Yes, I liked this suggestion. Gonna adopt it, thanks!

components/SocialButton/SocialButton.jsx
MutterPedro: :+1:

components/SocialButton/sub-components/GoogleIcon.unit.test.jsx
MutterPedro: :+1:

components/SocialButton/sub-components/FacebookIcon.unit.test.jsx
MutterPedro: :+1:

stories/Button/Button.story.jsx
rapahaeru: Good Question. We need to confirm this info with Arthur. 👍

components/SocialButton/SocialButton.jsx
```diff
@@ -0,0 +1,59 @@
...
+        <GoogleButton {...props} stroked skin="neutral">
+          <GoogleIcon title="google-button" size="24" />
+        </GoogleButton>
+      );
+    default:
+      return <Button />;
+  }
+};
+
+SocialButton.propTypes = {
```
rapahaeru: Thinking a little more, how about we put a size prop (not required) here too?
and that `size`, via this prop, passes to `GoogleIcon` and `FacebookIcon`?
That way, the size does not be in static way and who implements has this control.

Ps: we could keep the 24px size by default in Proptypes.

What do you think?

MutterPedro: I agree with you, that's the ideal. But I thought that it should come in other versions of the component since changing the icon size causes breaks in the button layout, so I think is better provide pre-defined sizes, like "small", "large", etc

Issue #130, opened 2019-09-11
Improvements to adding native lazy loading


tcelestino:
There's a new [attributes loading](https://web.dev/native-lazy-loading) implemented in new Chrome's version and can to use in the `<img>` and `<iframe>` tags. I've tried to add in the `<Card.Thumbnail loading='lazy' .../>` but this attribute doesn't show on component "compiled". 

Looking for code this component, I saw that use `{...props}` and it is created using to styled component. I don't that is a problem, because I tried to set others attributes and it didn't work too.

PR #211, opened 2020-03-24
feat(Card Thumbnail): implements native lazy load

rapahaeru:
## Description
task: https://jirasoftware.catho.com.br/browse/QTM-178
Issue: https://github.com/catho/quantum/issues/130

⚠️ this new attribute only will show in the  when the styled-components is updated.
In actual version `4.1.3`, this attribute did not exist and the styled removes the attributes that not recognise.
But, with this PR, when the styled-components is update, the attribute will show and works.

The problem in styled was solved here https://github.com/styled-components/styled-components/issues/2552

## Setup

### DIFF
--- a/components/Card/sub-components/Thumbnail.jsx
+++ b/components/Card/sub-components/Thumbnail.jsx
@@ -4,7 +4,7 @@ import styled from 'styled-components';
 
 import { colors } from '../../shared/theme';
 
-const Image = styled.img`
+const Image = styled.img.attrs({ loading: 'lazy' })`
   border-radius: ${({ rounded }) => (rounded ? '50%' : '4px')};
   display: inline-block;
   height: 72px;


components/Card/sub-components/Thumbnail.jsx
```diff
@@ -4,7 +4,7 @@ import styled from 'styled-components';
 
 import { colors } from '../../shared/theme';
 
-const Image = styled.img`
+const Image = styled.img.attrs({ loading: 'lazy' })`
```
renatogalvones: Doubt: Could this be solved by writing it inline like
```
const Thumbnail = ({ ...props }) => <Image {...props} loading="lazy" />;
```
?

rapahaeru: the effect is the same.
I tested it both ways...

renatogalvones: With this inline solution we still have to upgrade the styled components lib?

rapahaeru: unfortunately, yes!

PR #226, opened 2020-04-14
fix(Icons): make a default icon size on pre loading

rapahaeru:
## Description
image

This occurs because the css of the material loads after the render of svgs.
To avoid this, was putted a default size (0px) on pre loading process, for the svg icons doesnt break the layout.

⚠️ If your testing in catho-components, the icons still are loading in the wrong way, update the Quantum in node_modules of catho-components with the last build you made

## Setup
To simulate, build the Quantum and import in another application.

- [ ] Testing all the components in SSR (quantum-react-app)
- [ ] Testing some components in CSR (create-react-app)
- [ ] In application, import any Icon and simulate how its loading.

## Review guide
- [x] Regression \
- [x] Code review

### Browsers review
- [ ] Layout review
  - [ ] Edge
  - [ ] Firefox
  - [ ] Chrome

### DIFF
--- a/components/Icon/Icon.jsx
+++ b/components/Icon/Icon.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
 import AccessibleForward from '@material-ui/icons/AccessibleForward';
 import Add from '@material-ui/icons/Add';
 import ArrowDropDown from '@material-ui/icons/ArrowDropDown';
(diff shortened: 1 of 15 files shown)


components/Icon/Icon.jsx
```diff
@@ -116,7 +117,9 @@ const Icon = ({ name, skin, size, ...props }) => {
 
   if (!components[name]) return <span>{name}</span>;
 
-  const SelectedIcon = components[name];
+  const SelectedIcon = styled(components[name])`
+    width: 24px;
```
jgpacker, line 121:
```suggestion
    max-width: 24px;
```
Faz sentido fazer assim? Se sim, acho que a intenção pode ficar mais clara

tcelestino: @jgpacker, mas ai abre brecha para ter um icone com o tamanho menor, o que acredito que essa não seja a intenção. Correto @rapahaeru?

jgpacker: Pelo que entendi, o problema são ícones gigantes.

pedrosouza: I changed to initial width to zero because when I tested with different colors and sizes it was behaving oddly, starting with black than changing the color. Now the icon will appear only when the final style is loaded.

PR #227, opened 2020-04-27
feat(TabbedView): imeplements callback function on tab click

rapahaeru:
## Description
Add a new prop called `onTabClick` with a callback in onClick event of the tabs.
When some tab is clicked, the event will trigger

## Setup

## Review guide
      - `yarn test:coverage`
- [ ] Regression \
- [ ] Code review


components/TabbedView/TabbedView.jsx
```diff
@@ -182,8 +182,9 @@ class TabbedView extends React.Component {
     }
   }
 
-  onTabClick = tab => {
+  onTabClick = (tab, onTabClick) => {
```
tcelestino, line 185: What do you think to update type (ts) about component? @rapahaeru

PR #233, opened 2020-05-28
AUTH-298

luizjaccao:
## Description
Adds multi-line capacity to Tooltip component.

## Setup

## Review guide
      - `yarn test:coverage`
- [x] Regression \
- [ ] Code review

### Browsers review
- [ ] Layout review
  - [ ] Edge
  - [ ] IE 11
  - [ ] Firefox
  - [x] Chrome
  - [ ] Safari
  - [ ] Mobile
- [x] Ux review validation


luizjaccao:
Before:
image
After:
image

components/Tooltip/Tooltip.jsx
```diff
@@ -43,17 +43,18 @@ const Tip = styled.div`
 
 const TipText = styled.span`
   display: inline-block;
-  max-width: 250px;
```
luizjaccao, line 46: This line has been removed for the Tooltip component not to truncate text.

components/Tooltip/Tooltip.jsx
```diff
@@ -43,17 +43,18 @@ const Tip = styled.div`
...
   overflow: hidden;
   text-overflow: ellipsis;
-  white-space: nowrap;
+  white-space: pre;
```
luizjaccao, line 48: This line has been added for the Tootip component accept `\n` as enter.

components/Tooltip/Tooltip.jsx
```diff
@@ -43,17 +43,18 @@ const Tip = styled.div`
...
   line-height: 20px;
+  text-align: left;
```
luizjaccao, line 50: This line has been added to left align the text.

components/Tooltip/Tooltip.jsx
```diff
@@ -43,17 +43,18 @@ const Tip = styled.div`
...
 `;
 
 const Wrapper = styled.div`
   position: relative;
   float: left;
   clear: left;
+  width: max-content;
```
luizjaccao, line 57: This line has been added to adjust the size of Tooltip to the size of text.

components/Tooltip/Tooltip.jsx
tcelestino: what do you think to switch `max-width` to `min-width`? If you use it, you don't need remove this and you don't need to set `max-content` in line https://github.com/catho/quantum/pull/233/files#r431827523

components/Tooltip/Tooltip.jsx
tcelestino: what do you think to use `pre-wrap` instead?

components/Tooltip/Tooltip.jsx
tcelestino: what do you think to set this property using theme? That's can be used to _left_ or _center_ text.

components/Tooltip/Tooltip.jsx
```diff
@@ -12,6 +12,7 @@ const Tip = styled.div`
   border-radius: 4px;
   font-weight: bold;
   opacity: ${({ visible }) => (visible ? '1' : '0')};
+  visibility: ${({ visible }) => (visible ? 'visible' : 'hidden')};
```
luizjaccao, line 15: This line has been added to Tooltip only appear when mouse is over the wrapper.

components/Tooltip/Tooltip.jsx
luizjaccao: min-width change the expected layout 
image

components/Tooltip/Tooltip.jsx
luizjaccao: without `min-width` this change breaks the layout

components/Tooltip/Tooltip.jsx
pedrosouza: This change is breaking the layout:
image

Before the change:
image

pedrosouza: Why don't we add a prop to change the behavior for the case that you are addressing and keep the original behavior when not using this prop?

rapahaeru:
Suggestions:

- creates a new Boolean prop called `multiline` (example). If this props coming `true`, in the componente will have a condicional its implements the new needs. Keeping the original behavior when the prop was `false`.
-   when the new behavior was added, does not forget to implements its regression and unit tests for the new prop.

What do you think?

components/Tooltip/Tooltip.jsx
```diff
@@ -87,7 +91,7 @@ class Tooltip extends Component {
           visible={visibleProp || visibleState}
           theme={theme}
         >
-          <TipText>{text}</TipText>
+          <TipText multiline={multiline}>{text}</TipText>
```
rapahaeru, line 94: How about of creates a unit test for this new prop?

components/Tooltip/Tooltip.jsx
```diff
@@ -117,6 +122,7 @@ Tooltip.propTypes = {
 Tooltip.defaultProps = {
   placement: 'top',
   visible: false,
+  multiline: false,
```
rapahaeru, line 125: How about of create a new example, using this new prop, in story of storybook, for documentation?

rapahaeru, line 125: I think we should implement a new regression test using this new prop.

PR #333, opened 2021-08-06
fix(button): Adjusts component to supports icon only state

toncabral88:
## Description
Adjusts component to supports icon only state

## Setup

## Review guide
      - `yarn test:coverage`
- [ ] Regression \
- [ ] Code review

### Browsers review
- [ ] Layout review
  - [ ] Edge
  - [ ] Firefox
  - [ ] Chrome
  - [ ] Safari
  - [ ] Mobile


components/Icon/Icon.unit.test.jsx
```diff
@@ -19,6 +19,12 @@ describe('<Icon />', () => {
     const withLargeSize = renderer
       .create(<Icon name="search" size="large" />)
       .toJSON();
+    const withXSmallSize = renderer
+      .create(<Icon name="search" size="xlarge" />)
```
matheusrpp, line 23: I think the size for this was suposed to be "xsmall" and not "xlarge"? What do you think?

stories/Button/Button.regression-test.story.jsx
```diff
@@ -138,3 +138,32 @@ ErrorStroked.args = {
...
+export const IconOnlySmall = TemplateIconOnly.bind({});
+IconOnlySmall.args = {
+  size: 'small',
+};
+
+export const IconOnlyXSmall = TemplateIconOnly.bind({});
+IconOnlyXSmall.args = {
+  size: 'xsmall',
+};
+
```
matheusrpp, line 160: Since we are here, should we include a test for medium size to?

stories/Icon/Icon.regression-test.story.jsx
```diff
@@ -25,3 +25,13 @@ export const Small = Template.bind({});
 Small.args = {
   size: 'small',
 };
+
```
matheusrpp, line 28: And maybe another one here if we can. What do you think?

components/Button/Button.jsx
```diff
@@ -55,9 +57,17 @@ const StyledButton = styled.button`
   justify-content: center;
   border-radius: 4px;
 
-  ${props =>
+  ${({ iconOnly }) =>
```
matheusrpp, line 60: Question. Why not pass this props to ButtonIcon and then make a ternary to define the margin value? Right know it's just overwriting the old margin, isn't?  What do you think?

stories/Button/Button.regression-test.story.jsx
toncabral88: I didn't get it man, medium size is yhe default value.

components/Icon/Icon.unit.test.jsx
toncabral88: My bad!

stories/Button/Button.regression-test.story.jsx
matheusrpp: My bad, I was looking only what you changed in the tab files changed

stories/Button/Button.story.jsx
```diff
@@ -47,7 +47,11 @@ const exampleTab = (
           </LinkTo>
           .
         </p>
+
```
matheusrpp, line 50: I'm putting here just because of the file, but it would be nice to add a option to remove the text and show only the icon btn in the Usage tab. What do you think?

matheusrpp (changes requested):
## Review guide
      - `yarn test:coverage`
- [x] Regression \
- [ ] Code review

### Browsers review
- [x] Layout review
  - [x] Edge
  - [x] Firefox
  - [x] Chrome
  - [x] Safari
  - [x] Mobile

components/Button/Button.jsx
toncabral88: When I set iconOnly prop to ButtonIcon an error occurs and some components aren't rendered. I aggree that ButtonIcon shoud be reponsible by its styles, so to solve this I set a class to control this according button receives children.

stories/Button/Button.story.jsx
toncabral88: In this case, children prop receives a node or an array of nodes and Quantum Ui does not support edition for these props.

rapahaeru (approved):
## Review guide
      - `yarn test:coverage`
- [x] Regression \
- [x] Code review

### Browsers review
- [x] Layout review

PR #363, opened 2022-01-12
fix(QTM-321): Adjusts on ts props of all components

toncabral88:
## Description
https://jirasoftware.catho.com.br/browse/QTM-321

## Setup

## Review guide
- [ ] Code review
- [ ] TypeScript updated

### DIFF
--- a/components/Accordion/index.d.ts
+++ b/components/Accordion/index.d.ts
@@ -2,8 +2,8 @@ import React from 'react';
 
 export interface AccordionProps {
     theme?: {
-        colors?: object;
-        spacing?: object;
+        colors?: {};
+        spacing?: {};
         baseFontSize?: number;
     };
     keepOnlyOneOpen?: boolean;
--- a/components/Avatar/index.d.ts
+++ b/components/Avatar/index.d.ts
@@ -6,7 +6,7 @@ export interface AvatarProps {
     text?: string;
     hasToggle?: boolean;
     href?: string;
-    onClick?: () => {};
+    onClick?: () => void;
 }
 
 export default class Avatar extends React.Component<AvatarProps> { }
--- a/components/Avatar/index.d.ts
+++ b/components/Avatar/index.d.ts
@@ -6,7 +6,6 @@ export interface AvatarProps {
     text?: string;
     hasToggle?: boolean;
     href?: string;
-    size?: string;
     onClick?: () => {};
 }
 
--- a/components/Button/index.d.ts
+++ b/components/Button/index.d.ts
@@ -23,7 +23,6 @@ export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
             button: {};
         };
     };
-    id?: string;
 }
 
 export default class Button extends React.Component<ButtonProps> {
--- a/components/Button/index.d.ts
+++ b/components/Button/index.d.ts
@@ -1,12 +1,13 @@
-import React from 'react';
+import React, { ButtonHTMLAttributes } from 'react';
+import { IconNames } from '../Icon';
 import IconButton from './IconButton';
 
-export interface ButtonProps {
+export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
     center?: boolean;
     disabled?: boolean;
     stroked?: boolean;
     full?: boolean;
-    icon?: string;
+    icon?: IconNames | string;
     size?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge';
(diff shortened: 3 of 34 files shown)


components/Avatar/index.d.ts
```diff
@@ -6,7 +6,6 @@ export interface AvatarProps {
     text?: string;
     hasToggle?: boolean;
     href?: string;
-    size?: string;
     onClick?: () => {};
```
matheusrpp, line 9: Here the onClick as prop is declared only in the index file. Is supposed to be like this?

matheusrpp (changes requested):
Hey, a lot of this comments are more questions and doubts than changes. Some of then I didn't know if it was supposed like that ore not.

In the component SocialButton, shouldn't be there a onClick in the index.d.ts?

## Review guide
- [ ] Code review
- [ ] TypeScript updated

components/Input/index.d.ts
```diff
@@ -32,12 +31,12 @@ export interface InputProps {
   mask?: Mask | Mask[];
```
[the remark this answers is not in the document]
toncabral88: I think this is here to support validation props for form.

components/Toggle/index.d.ts
```diff
@@ -5,8 +5,8 @@ export interface ToggleProps {
   id?: string;
```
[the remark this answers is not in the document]
toncabral88: Yes, the id is here because the rest is declared on component.

components/Button/index.d.ts
```diff
@@ -15,11 +16,11 @@ export interface ButtonProps {
...
-        breakpoints?: object;
+        colors?: {};
+        spacing?: {};
+        breakpoints?: {};
         components?: {
-            button: object;
+            button: {};
         };
     };
     id?: string;
```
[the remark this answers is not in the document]
toncabral88: In this case isn't necessary because the button extends HTML button atributes.

components/Avatar/index.d.ts
toncabral88: Same case of toggle here.

components/Accordion/index.d.ts
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 
 export interface AccordionProps {
     theme?: {
-        colors?: object;
-        spacing?: object;
+        colors?: {};
+        spacing?: {};
```
tcelestino, line 6: You could use `Record<>`  instead of `{}` - https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

```suggestion
        colors?: Record<string, unknown>;
        spacing?: Record<string, unknown>;
```

components/Avatar/index.d.ts
tcelestino, line 9:
```suggestion
    onClick?: () => void;
```

components/RadioGroup/index.d.ts
```diff
@@ -22,17 +23,17 @@ export interface RadioButtonProps {
   size?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge';
   disabled?: boolean;
   error?: boolean;
-  icon?: string;
+  icon?: IconNames | string;
   id?: string;
   inline?: boolean;
   label?: string;
-  onChange?: React.ChangeEventHandler<HTMLInputElement>;
+  onChange?: () => {} | React.ChangeEventHandler<HTMLInputElement>;
```
tcelestino, line 30:
```suggestion
  onChange?: () => void | React.ChangeEventHandler<HTMLInputElement>;
```

components/TabbedView/index.d.ts
```diff
@@ -1,28 +1,30 @@
...
+            tabbedView?: {};
         };
         baseFontSize?: number;
-        breakpoints?: object;
-        spacing?: object;
+        breakpoints?: {};
+        spacing?: {};
     };
-    onClick?: () => void;
+    onTabClick?: () => {};
```
tcelestino, line 25: why not `void`?

tcelestino (dismissed):
- [x] Code review
- [x] TypeScript updated

I added suggestions

components/Accordion/index.d.ts
toncabral88: Man, this is interesting. I would like to create an interface to  theme items and include it on ts definitions, but we're considering set a ThemeProvider and remove theme from all components. I don't know if worth the effort.

tcelestino: No problem! If you'll use ThemeProvider, this is deprecated soon. 😎

tcelestino (approved):
LGTM

PR #370, opened 2022-02-01
Tests(Qtm-325) - Migrating Circular Loader, Dialog and Dropdown Unit tests

matheusrpp:
## Description
https://jirasoftware.catho.com.br/browse/QTM-325

## Review guide
      - `yarn test:coverage`
- [ ] Code review


components/Dialog/Dialog.unit.test.jsx
```diff
@@ -54,7 +47,7 @@ describe('Dialog', () => {
     });
 
     const onCloseMock = jest.fn();
-    mount(<Dialog onClose={onCloseMock} />);
+    render(<Dialog onClose={onCloseMock} />);
```
toncabral88, line 50: Here is missing a test when another key is pressed. My suggestions is something like:

`
eventMap.keydown({ key: 'Enter' });
eventMap.keydown({ key: 'Escape' });

expect(onCloseMock).toHaveBeenCalledTimes(1);
`

or another test to check if function is called when another key is pressed.

toncabral88 (changes requested):
Coverage is the same, but I think you can improve some tests like these lines of dropdown:

image

image

image

I don't know how it's hard, but I only request the dialog change, in the dropdown could be nice coverage these suggestions.

components/Dialog/Dialog.unit.test.jsx
matheusrpp: About the dropdown, didn't manage to cover all this scenarios, but covered a few more.

toncabral88 (approved):
      - `yarn test:coverage`
- [x] Code review

PR #444, opened 2022-10-11
docs: (Popover) QTM-380 Updated documentation template

douglasjr97:
## Description
https://jirasoftware.catho.com.br/browse/QTM-380

## Review guide
- [ ] Code review
- [ ] Documentation


stories/Popover/Popover.stories.jsx
```diff
@@ -0,0 +1,201 @@
...
+  skin: 'neutral',
+  children: 'childrenText...',
+  trigger: 'triggerText...',
+};
+
+export const PrimaryTop = Template.bind({});
+PrimaryTop.args = {
+  direction: 'top',
+  placement: 'top',
+  skin: 'primary',
```
MarcosViniciusPC, line 63: In the same way that "placement" happens here with the "skin". If there is already an example of using the prop skin, you don't need other examples for all possible "skin" values

stories/Popover/Popover.stories.jsx
```diff
@@ -0,0 +1,201 @@
+import Popover from '../../components/Popover';
+import { Row, Col } from '../../components';
+
+const Template = args => {
+  const { children: Component, ...rest } = args;
+  return (
+    <>
+      <Row style={{ paddingTop: '72px', paddingLeft: '414px' }}>
```
MarcosViniciusPC, line 8: We can follow the stylization adopted in the tooltip documentation in this case of not centering the popover using paddingLeft

MarcosViniciusPC (changes requested):
I have described some situations below that can be improved. Also, I believe it is necessary to include new examples of the popover with the props "inverted" and "visible". You can check examples of updated documentation to understand how we seek to exemplify the possible states of each component

stories/Popover/Popover.stories.jsx
```diff
@@ -0,0 +1,201 @@
...
+
+export const Default = Template.bind({});
+Default.args = {
+  children: 'childrenText...',
+  trigger: 'triggerText...',
+};
+
+export const NeutralTop = Template.bind({});
+NeutralTop.args = {
+  direction: 'top',
```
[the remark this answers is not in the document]
douglasjr97: Direction existed in popover Popover.regression-test.story.jsx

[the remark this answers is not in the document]
MarcosViniciusPC: The "direction" argument is passed to the "placement" property in the popover component's regression tests template. I recommend not doing this. I find it much more interesting to check and pass the properties of each component as defined propTypes.

stories/Popover/Popover.stories.jsx
```diff
@@ -0,0 +1,42 @@
...
+  const { ...rest } = args;
+  return (
+    <Row
+      style={{
+        padding: '44px 0',
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'center',
+        alignItems: 'center',
+      }}
```
MarcosViniciusPC, line 14: Just a padding of 72px I believe is enough

stories/Popover/Popover.stories.jsx
```diff
@@ -0,0 +1,42 @@
...
+        <Popover trigger="TriggerText" {...rest}>
+          childrenText
+        </Popover>
+      </Col>
+    </Row>
+  );
+};
+
+export const Default = Template.bind({});
+Default.args = {};
```
MarcosViniciusPC, line 26:
```suggestion
```
When an example has no parameters it is not necessary to include an "args" object

stories/Popover/Popover.stories.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const WithSkin = Template.bind({});
+WithSkin.args = {
+  placement: 'top',
```
MarcosViniciusPC, line 30: I see no need to include this parameter in this scenario

MarcosViniciusPC (changes requested):
In addition to the points mentioned below, I also think it's necessary to include an example using the "visible" property

MarcosViniciusPC (approved):
[x] Code review
[x] Documentation

PR #454, opened 2022-10-25
docs: (Socials) QTM-390 updated documentation template

douglasjr97:
## Description
https://jirasoftware.catho.com.br/browse/QTM-390

## Review guide
- [ ] Code review
- [ ] Documentation


stories/Socials/Socials.stories.jsx
```diff
@@ -0,0 +1,45 @@
...
+    name: 'youtube',
+    url: 'https://youtube.com',
+    description: 'youtube description of the purpose of this icon',
+  },
+];
+
+export const AllItems = Template.bind({});
+AllItems.args = {
+  items: allItems,
+};
```
MarcosViniciusPC, line 26:
```suggestion
```
When putting a default value in the template, this argument is not needed in this example

stories/Socials/Socials.stories.jsx
```diff
@@ -0,0 +1,45 @@
...
+export const AllItemsWithoutBox = Template.bind({});
+
+AllItemsWithoutBox.args = {
+  ...AllItems.args,
+  withBox: false,
+};
+
+export const OneItemWithouBox = Template.bind({});
+OneItemWithouBox.args = {
+  items: [allItems[1]],
```
MarcosViniciusPC, line 37:
```suggestion
```
This prop would not be necessary in this example

MarcosViniciusPC (changes requested):
In addition to the suggestions below, is there a need to keep the "stories/Socials/examples" folder? What do you think about removing it?

stories/Socials/Socials.stories.jsx
```diff
@@ -0,0 +1,45 @@
...
+];
+
+export const AllItems = Template.bind({});
+AllItems.args = {
+  items: allItems,
+};
+
+export const AllItemsWithoutBox = Template.bind({});
+
+AllItemsWithoutBox.args = {
```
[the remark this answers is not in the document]
douglasjr97: I don't think this name would be good because the pattern already comes with all the items. the difference with this property is that it is not inside a box

MarcosViniciusPC (approved):
[x] Code review
[x] Documentation

PR #466, opened 2022-12-28
build: (QTM-423) - updated node and environment files in pipeline

douglasjr97:
## Description
https://jirasoftware.catho.com.br/browse/qtm-423

## Review guide
- [ ] Code review

### DIFF
--- a/.github/workflows/actions-development.yml
+++ b/.github/workflows/actions-development.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Read .nvmrc
-        run: echo NVMRC={cat .nvmrc} >> $GITHUB_OUTPUT
+        run: echo NVMRC=$(cat .nvmrc) >> $GITHUB_OUTPUT
         id: nvm
 
       - name: Setup node
--- a/.github/workflows/actions-development.yml
+++ b/.github/workflows/actions-development.yml
@@ -7,13 +7,13 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Read .nvmrc
         run: echo ::set-output name=NVMRC::$(cat .nvmrc)
         id: nvm
 
       - name: Setup node
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
           node-version: '${{ steps.nvm.outputs.NVMRC }}'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/actions-development.yml
```diff
@@ -7,13 +7,13 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Read .nvmrc
-        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
+        run: echo {NVMRC}={cat .nvmrc} >> $GITHUB_OUTPUT
```
MarcosViniciusPC, line 12: This command is wrong, notice that NVRMC is being defined as value instead of variable. Due to this problem, the node version used in the pipeline is different from the node version in the .NVMRC file.

Maybe the suggestion below will work

```suggestion
        run: echo NVMRC={cat .nvmrc} >> $GITHUB_OUTPUT
```

.github/workflows/actions-publish.yml
```diff
@@ -10,13 +10,13 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Read .nvmrc
-        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
+        run: echo {NVMRC}={cat .nvmrc} >> $GITHUB_OUTPUT
```
MarcosViniciusPC, line 15: This command is wrong, notice that NVRMC is being defined as value instead of variable. Due to this problem, the node version used in the pipeline is different from the node version in the .NVMRC file.

Maybe the suggestion below will work

```suggestion
        run: echo NVMRC={cat .nvmrc} >> $GITHUB_OUTPUT
```

.github/workflows/actions-publish.yml
```diff
@@ -37,13 +37,13 @@ jobs:
...
```
MarcosViniciusPC, line 42: This command is wrong, notice that NVRMC is being defined as value instead of variable. Due to this problem, the node version used in the pipeline is different from the node version in the .NVMRC file.

Maybe the suggestion below will work
```suggestion
        run: echo NVMRC={cat .nvmrc} >> $GITHUB_OUTPUT
```

MarcosViniciusPC (changes requested):
[] Core review

MarcosViniciusPC (approved):
[x] Code review

PR #474, opened 2023-02-16
perf: (QTM-444) preload on font loading

douglasjr97:
## Description
https://jirasoftware.catho.com.br/browse/QTM-444

## Review guide
- [ ] Regression
- [ ] Code review
- [ ] Documentation

### Browsers review
- [ ] Layout review
  - [ ] Edge
  - [ ] Firefox
  - [ ] Chrome
  - [ ] Safari
  - [ ] Mobile

### DIFF
--- a/components/GlobalStyle/sub-components/Typography.jsx
+++ b/components/GlobalStyle/sub-components/Typography.jsx
@@ -94,7 +94,7 @@ const Typography = () => (
       crossOrigin="anonymous"
       rel="preload"
       href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,600,700,400i,600i,700i&display=swap"
-      as="font"
+      as="style"
     />
     <link
       href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,600,700,400i,600i,700i&display=swap"
--- a/components/GlobalStyle/sub-components/Typography.jsx
+++ b/components/GlobalStyle/sub-components/Typography.jsx
@@ -90,17 +90,21 @@ ul, ol {
 const Typography = () => (
   <>
     <Style />
+
     <link
+      rel="preconnect"
+      href="https://fonts.gstatic.com"
       crossOrigin="anonymous"
+    />
+    <link
       rel="preload"
       href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,600,700,400i,600i,700i&display=swap"
       as="font"
-      type="font/woff2"
+      crossOrigin="anonymous"
     />
     <link
       href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,600,700,400i,600i,700i&display=swap"
       rel="stylesheet"
-      as="style"
     />
   </>
 );
--- a/components/GlobalStyle/sub-components/Typography.jsx
+++ b/components/GlobalStyle/sub-components/Typography.jsx
@@ -94,11 +94,13 @@ const Typography = () => (
       crossOrigin="anonymous"
       rel="preload"
       href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,600,700,400i,600i,700i&display=swap"
-      as="style"
+      as="font"
+      type="font/woff2"
     />
     <link
       href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,600,700,400i,600i,700i&display=swap"
       rel="stylesheet"
+      as="style"
     />
   </>
 );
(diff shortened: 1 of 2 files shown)


MarcosViniciusPC (changes requested):
[] Code review
[] Regression

I assume that the preload in this case caused the google font not to be applied to the quantum typography and this changed all the project screenshots. Try to understand more about preload behavior in mozilla documents.

MarcosViniciusPC (commented):
And again, try to understand how the resources you use work. Copy and paste, but seek to understand, reflect. One way to do this is by consulting official documents and the main references on the subject.

douglasjr97:
> And again, try to understand how the resources you use work. Copy and paste, but seek to understand, reflect. One way to do this is by consulting official documents and the main references on the subject.

I trusted these sources
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload
https://pt.stackoverflow.com/questions/269556/o-que-%C3%A9-rel-preload

components/GlobalStyle/sub-components/Typography.jsx
```diff
@@ -90,9 +90,17 @@ ul, ol {
...
+      crossOrigin="anonymous"
+      rel="preload"
+      href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,600,700,400i,600i,700i&display=swap"
+      as="font"
+      type="font/woff2"
+    />
     <link
       href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,600,700,400i,600i,700i&display=swap"
       rel="stylesheet"
+      as="style"
```
MarcosViniciusPC, line 103: https://developer.mozilla.org/pt-BR/docs/Web/HTML/Element/link#attr-as
I want you to read the documentation calmly, as I instructed you. Always look for content from great references on the internet. There are many people on the internet who use some resources erroneously.

components/GlobalStyle/sub-components/Typography.jsx
```diff
@@ -90,9 +90,17 @@ ul, ol {
 const Typography = () => (
   <>
     <Style />
+    <link
+      crossOrigin="anonymous"
+      rel="preload"
+      href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,600,700,400i,600i,700i&display=swap"
+      as="font"
+      type="font/woff2"
```
MarcosViniciusPC, line 98: You point out that the style is a font, but the href points to a css. This is divergent. I believe you are relying too much on the referral links you sent in slack. Note that in the mentioned links the problem they want to solve is importing a font and not importing a css

components/GlobalStyle/sub-components/Typography.jsx
```diff
@@ -90,9 +90,17 @@ ul, ol {
 const Typography = () => (
   <>
     <Style />
+    <link
+      crossOrigin="anonymous"
+      rel="preload"
+      href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,600,700,400i,600i,700i&display=swap"
+      as="font"
```
MarcosViniciusPC, line 97: Divergence with the href

MarcosViniciusPC (changes requested):
[] Code review

components/GlobalStyle/sub-components/Typography.jsx
douglasjr97:
```suggestion
      type="font/woff2"
```
That's why I added this format, with its presence the font extension is inside the css file, understand?

components/GlobalStyle/sub-components/Typography.jsx
douglasjr97: I checked in the documentation that the "from" attribute is only used on links that use (preload).

components/GlobalStyle/sub-components/Typography.jsx
douglasjr97: The Font Attribute already references the CSS/@font-face type, which is used in this case for the google font.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-type

image

components/GlobalStyle/sub-components/Typography.jsx
douglasjr97: I evaluated that when we use "as=Font" there is no need to use a typing

components/GlobalStyle/sub-components/Typography.jsx
MarcosViniciusPC: Very good!

MarcosViniciusPC (approved):
[x] Code review

rapahaeru (approved):

PR #481, opened 2023-03-06
(Input/modal) QTM-442 - Problems with compounded components and intellisense

douglasjr97:
## Description
https://jirasoftware.catho.com.br/browse/QTM-442

## Review guide
- [ ] Regression
- [ ] Code review
- [ ] Documentation
- [ ] Build

### Browsers review
- [ ] Layout review
  - [ ] Edge
  - [ ] Firefox
  - [ ] Chrome
  - [ ] Safari
  - [ ] Mobile

Script for tests:

1. Yarn Build (QTM-442)
2. Choose some other project that uses typescript
3. Open the package.json file
4. Change the Quantum CDN path to the location pointing to the Quantum build folder
image

### DIFF
--- a/components/Input/index.d.ts
+++ b/components/Input/index.d.ts
@@ -1,4 +1,7 @@
 import { InputHTMLAttributes, MouseEventHandler, ChangeEventHandler } from 'react';
+import masks from '../shared/masks';
+import Input from './Input';
+
 
 import { Validations } from '../Form';
 
@@ -41,13 +44,14 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
   };
 }
 
+
 interface CompoundedInput extends React.FunctionComponent<InputProps> {
-  CEP: React.FunctionComponent;
-  CNPJ: React.FunctionComponent;
-  CPF: React.FunctionComponent;
-  Date: React.FunctionComponent;
-  Password: React.FunctionComponent;
-  Phone: React.FunctionComponent;
+  CEP: React.FunctionComponent<InputProps>;
+  CNPJ: React.FunctionComponent<InputProps>;
+  CPF: React.FunctionComponent<InputProps>;
+  Date: React.FunctionComponent<InputProps>;
+  Password: React.FunctionComponent<InputProps>;
+  Phone: React.FunctionComponent<InputProps>;
 }
 
 const Input: CompoundedInput = () => null
--- a/components/Input/index.d.ts
+++ b/components/Input/index.d.ts
@@ -1,4 +1,4 @@
-import { Component, InputHTMLAttributes, MouseEventHandler, ChangeEventHandler } from 'react';
+import { InputHTMLAttributes, MouseEventHandler, ChangeEventHandler } from 'react';
 
 import { Validations } from '../Form';
 
@@ -41,4 +41,15 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
   };
 }
 
-export default class Input extends Component<InputProps> {}
+interface CompoundedInput extends React.FunctionComponent<InputProps> {
+  CEP: React.FunctionComponent<InputProps>;
+  CNPJ: React.FunctionComponent<InputProps>;
+  CPF: React.FunctionComponent<InputProps>;
+  Date: React.FunctionComponent<InputProps>;
+  Password: React.FunctionComponent<InputProps>;
+  Phone: React.FunctionComponent<InputProps>;
+}
+
+const Input: CompoundedInput = () => null
+
+export default Input;
--- a/components/Input/index.d.ts
+++ b/components/Input/index.d.ts
@@ -1,8 +1,4 @@
 import { InputHTMLAttributes, MouseEventHandler, ChangeEventHandler } from 'react';
-import masks from '../shared/masks';
-import Input from './Input';
-
-
 import { Validations } from '../Form';
 
 export type MaskFunction = (rawValue: string) => string[];
--- a/components/Input/index.d.ts
+++ b/components/Input/index.d.ts
@@ -42,12 +42,12 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
 }
 
 interface CompoundedInput extends React.FunctionComponent<InputProps> {
-  CEP: React.FunctionComponent<InputProps>;
-  CNPJ: React.FunctionComponent<InputProps>;
-  CPF: React.FunctionComponent<InputProps>;
-  Date: React.FunctionComponent<InputProps>;
-  Password: React.FunctionComponent<InputProps>;
-  Phone: React.FunctionComponent<InputProps>;
+  CEP: React.FunctionComponent;
+  CNPJ: React.FunctionComponent;
+  CPF: React.FunctionComponent;
+  Date: React.FunctionComponent;
+  Password: React.FunctionComponent;
+  Phone: React.FunctionComponent;
 }
 
 const Input: CompoundedInput = () => null
--- a/components/Modal/index.d.ts
+++ b/components/Modal/index.d.ts
@@ -1,13 +1,4 @@
-import { ComponentType, ReactNode, TouchEvent, MouseEventHandler } from 'react';
-
-interface StaticProps {
-  theme?: {
-    spacing?: {};
-  };
-}
-
-export type ModalHeader = ComponentType<StaticProps>;
-export type ModalContent = ComponentType<StaticProps>;
+import { ReactNode, TouchEvent, MouseEventHandler } from 'react';
 
 export interface ModalProps {
   children?: ReactNode[] | ReactNode;
@@ -24,7 +15,6 @@ export interface ModalProps {
   };
 }
 
-  
   interface ModalHeaderProps {
     theme?: {
       spacing?: {};
@@ -50,9 +40,6 @@ interface ModalFooterProps {
     spacing?: {};
   };
 }
-
-
-
 
 interface CompoundedModal extends React.FunctionComponent<ModalProps> {
   Header: React.FunctionComponent<ModalHeaderProps>;
--- a/components/Modal/index.d.ts
+++ b/components/Modal/index.d.ts
@@ -24,12 +24,42 @@ export interface ModalProps {
   };
 }
 
+  
+  interface ModalHeaderProps {
+    theme?: {
+      spacing?: {};
+    };
+  }
+
+  interface ModalTitleProps {
+    theme?: { 
+      baseFontSize?: {}; 
+    };
+}
+
+
+interface ModalContentProps {
+  theme?: { 
+    baseFontSize?: number;
+     spacing?: {} 
+    };
+}
+
+interface ModalFooterProps {
+  theme?: {
+    spacing?: {};
+  };
+}
+
+
+
+
 interface CompoundedModal extends React.FunctionComponent<ModalProps> {
-  Header: React.FunctionComponent<ModalProps>;
-  Title: React.FunctionComponent<ModalProps>;
-  HeaderText: React.FunctionComponent<ModalProps>;
-  Content: React.FunctionComponent<ModalProps>;
-  Footer: React.FunctionComponent<ModalProps>;
+  Header: React.FunctionComponent<ModalHeaderProps>;
+  Title: React.FunctionComponent<ModalTitleProps>;
+  HeaderText: React.FunctionComponent;
+  Content: React.FunctionComponent<ModalContentProps>;
+  Footer: React.FunctionComponent<ModalFooterProps>;
 }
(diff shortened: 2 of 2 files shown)


components/Modal/index.d.ts
```diff
@@ -27,14 +24,14 @@ export interface ModalProps {
...
+interface CompoundedModal extends React.FunctionComponent<ModalProps> {
+  Header: React.FunctionComponent<ModalProps>;
+  Title: React.FunctionComponent<ModalProps>;
+  HeaderText: React.FunctionComponent<ModalProps>;
+  Content: React.FunctionComponent<ModalProps>;
+  Footer: React.FunctionComponent<ModalProps>;
+}
 
-  static Content: ModalContent;
+const Modal: CompoundedModal = () => null;
```
MarcosViniciusPC, line 35:
```suggestion
```
Don't constantly cry on type file

components/Modal/index.d.ts
```diff
@@ -27,14 +24,14 @@ export interface ModalProps {
   };
 }
 
-export default class Modal extends Component<ModalProps> {
-  static Header: ModalHeader;
-
-  static Title: Title;
-
-  static HeaderText: HeaderText;
+interface CompoundedModal extends React.FunctionComponent<ModalProps> {
```
MarcosViniciusPC, line 27:
```suggestion
interface Modal extends React.Component<ModalProps> {
```
Modal is not a functional component

components/Modal/index.d.ts
```diff
@@ -1,7 +1,4 @@
-import { Component, ComponentType, ReactNode, TouchEvent, MouseEventHandler } from 'react';
-import HeaderText from '../Card/HeaderText';
-import Title from '../Card/Title';
-import Footer from '../Card/Footer';
+import { ComponentType, ReactNode, TouchEvent, MouseEventHandler } from 'react';
```
MarcosViniciusPC, line 1:
```suggestion
import { Component, ComponentType, ReactNode, TouchEvent, MouseEventHandler } from 'react';
```

components/Modal/index.d.ts
```diff
@@ -27,14 +24,14 @@ export interface ModalProps {
   };
 }
 
-export default class Modal extends Component<ModalProps> {
-  static Header: ModalHeader;
```
MarcosViniciusPC, line 31: Is "ModalHeader" still used in some scenario? If not, remove the type

components/Modal/index.d.ts
```diff
@@ -27,14 +24,14 @@ export interface ModalProps {
...
-
-  static Title: Title;
```
MarcosViniciusPC, line 33: Is "Title" still used in some scenario? If not, remove the type

components/Modal/index.d.ts
```diff
@@ -27,14 +24,14 @@ export interface ModalProps {
...
-
-  static HeaderText: HeaderText;
```
MarcosViniciusPC, line 35: Is "ModalText" still used in some scenario? If not, remove the type

components/Modal/index.d.ts
```diff
@@ -27,14 +24,14 @@ export interface ModalProps {
...
+interface CompoundedModal extends React.FunctionComponent<ModalProps> {
+  Header: React.FunctionComponent<ModalProps>;
+  Title: React.FunctionComponent<ModalProps>;
+  HeaderText: React.FunctionComponent<ModalProps>;
+  Content: React.FunctionComponent<ModalProps>;
+  Footer: React.FunctionComponent<ModalProps>;
```
MarcosViniciusPC, line 32: Have you verified that all types of these subcomponents are functional components? I do not think so

MarcosViniciusPC: Do these subcomponents have the same properties as the parent component? I do not think so

components/Input/index.d.ts
```diff
@@ -41,4 +41,15 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
   };
 }
 
-export default class Input extends Component<InputProps> {}
+interface CompoundedInput extends React.FunctionComponent<InputProps> {
+  CEP: React.FunctionComponent<InputProps>;
+  CNPJ: React.FunctionComponent<InputProps>;
+  CPF: React.FunctionComponent<InputProps>;
+  Date: React.FunctionComponent<InputProps>;
```
MarcosViniciusPC, line 48: How about saying that this subcomponents has all the properties of the Input component except the "mask" property?

components/Input/index.d.ts
```diff
@@ -41,4 +41,15 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
...
+  Password: React.FunctionComponent<InputProps>;
+  Phone: React.FunctionComponent<InputProps>;
```
MarcosViniciusPC, line 50: How about saying that this subcomponent has all the properties of the Input component except the "mask" property?

components/Input/index.d.ts
```diff
@@ -41,4 +41,15 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
   };
 }
 
-export default class Input extends Component<InputProps> {}
+interface CompoundedInput extends React.FunctionComponent<InputProps> {
+  CEP: React.FunctionComponent<InputProps>;
+  CNPJ: React.FunctionComponent<InputProps>;
+  CPF: React.FunctionComponent<InputProps>;
+  Date: React.FunctionComponent<InputProps>;
+  Password: React.FunctionComponent<InputProps>;
```
MarcosViniciusPC, line 49: How about saying that this subcomponent has all the properties of the Input component except the "type" property?

components/Input/index.d.ts
```diff
@@ -41,4 +41,15 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
   };
 }
 
-export default class Input extends Component<InputProps> {}
+interface CompoundedInput extends React.FunctionComponent<InputProps> {
```
MarcosViniciusPC, line 44:
```suggestion
interface CompoundedInput extends React.Component<InputProps> {
```
Input is not a functional component

components/Input/index.d.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, InputHTMLAttributes, MouseEventHandler, ChangeEventHandler } from 'react';
+import { InputHTMLAttributes, MouseEventHandler, ChangeEventHandler } from 'react';
```
MarcosViniciusPC, line 1:
```suggestion
import { Component, InputHTMLAttributes, MouseEventHandler, ChangeEventHandler } from 'react';
```

components/Input/index.d.ts
```diff
@@ -41,4 +41,15 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
...
+interface CompoundedInput extends React.FunctionComponent<InputProps> {
+  CEP: React.FunctionComponent<InputProps>;
+  CNPJ: React.FunctionComponent<InputProps>;
+  CPF: React.FunctionComponent<InputProps>;
+  Date: React.FunctionComponent<InputProps>;
+  Password: React.FunctionComponent<InputProps>;
+  Phone: React.FunctionComponent<InputProps>;
+}
+
+const Input: CompoundedInput = () => null
```
MarcosViniciusPC, line 53:
```suggestion
```
Don't constantly cry on type file

components/Input/index.d.ts
MarcosViniciusPC, line 44:
```suggestion
interface Input extends React.Component<InputProps> {
```

MarcosViniciusPC (changes requested):
[] Code review
[x] Build

Test apps

[x] Job-search 
[x] ssr-simulator

In addition to the points noted below, it might be interesting to include the subcomponents in the proptypes of the parent component in order to show them in the documentation, what do you think?

components/Modal/index.d.ts
douglasjr97: React.Component doesn't work with subcomponents, Because it is not possible to type in internal components.

components/Modal/index.d.ts
douglasjr97: React.Component doesn't work with subcomponents, Because it is not possible to type in internal components.

components/Modal/index.d.ts
douglasjr97: i agree with a remotion

components/Modal/index.d.ts
douglasjr97: You are pointing to the version that hasn't changed yet

components/Modal/index.d.ts
douglasjr97: You are pointing to the version that hasn't changed yet

components/Modal/index.d.ts
douglasjr97: If subcomponents are used as class instead of FunctionalComponents the problem will continue

components/Modal/index.d.ts
MarcosViniciusPC: Why? Can you show an attachment with the problem?

components/Modal/index.d.ts
MarcosViniciusPC: Why? Can you show an attachment with the problem?

components/Modal/index.d.ts
MarcosViniciusPC: ??

components/Input/index.d.ts
douglasjr97: it's not necessary

components/Input/index.d.ts
douglasjr97: it's not necessary

components/Input/index.d.ts
douglasjr97: cannot adjust children if parent component is a class component

components/Input/index.d.ts
douglasjr97: cannot adjust children if parent component is a class component

components/Modal/index.d.ts
douglasjr97: image-2023-01-27-16-40-39-830

components/Modal/index.d.ts
douglasjr97: this treatment was necessary to "convert" the class component into a functional one

MarcosViniciusPC (changes requested):
In intellisense, all properties of the parent component are shown in the subcomponents, but not always the subcomponents have the same properties as the parent. Example: Modal component has an attribute called onClose, but Modal.Content does not, but it is suggested by intellissense.

MarcosViniciusPC (changes requested):
[] Code review

In addition to the points mentioned below, I had problems testing today, has the same situation happened to you?
image

components/Input/index.d.ts
```diff
@@ -41,4 +41,15 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
...
 }
 
-export default class Input extends Component<InputProps> {}
+interface CompoundedInput extends React.FunctionComponent<InputProps> {
+  CEP: React.FunctionComponent;
+  CNPJ: React.FunctionComponent;
+  CPF: React.FunctionComponent;
+  Date: React.FunctionComponent;
+  Password: React.FunctionComponent;
+  Phone: React.FunctionComponent;
```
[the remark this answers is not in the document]
douglasjr97: The ideal would be to go back to the way it was before, receiving the father's properties.

MarcosViniciusPC (approved):
[x] Code review
[x] Build

components/Input/index.d.ts
```diff
@@ -1,4 +1,7 @@
-import { Component, InputHTMLAttributes, MouseEventHandler, ChangeEventHandler } from 'react';
+import { InputHTMLAttributes, MouseEventHandler, ChangeEventHandler } from 'react';
+import masks from '../shared/masks';
```
rapahaeru, line 2: Where is this import used?

components/Input/index.d.ts
```diff
@@ -1,4 +1,7 @@
...
+import Input from './Input';
```
rapahaeru, line 3: Where is this import used?

components/Modal/index.d.ts
```diff
@@ -27,14 +15,40 @@ export interface ModalProps {
...
+    theme?: {
+      spacing?: {};
+    };
+  }
+
+  interface ModalTitleProps {
+    theme?: { 
+      baseFontSize?: {}; 
+    };
+}
```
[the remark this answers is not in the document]
douglasjr97: i did add the remaining property

jonatasprates (approved):

rapahaeru (approved):

PR #488, opened 2023-04-07
Refactor Skeleton component to introduce 'as' prop for polymorphic render support

### LINKED ISSUE #437: Skeleton.Text
### Skeleton.Text html structure

when using the Skeleton.Text component it keeps logging this warning about HTML semantic issue:
image

### Proposed solution
Refactor the Skeleton.Text component to use span as a generic inline element will not trigger this problem and because the base Skeleton uses a div with `display: inline-block` it will not have any problems due to this change

### Our Context
We use the Skeleton.Text inside the paragraph that contains a label and then the skelenton for the data that is displayed after an API call:
image
With the solution proposed it would make the HTML semantic and possible to use it inside a `<p></p>` or a `<div></div>`

2bfe26:
- Closes #437

This PR addresses an issue with the Skeleton.Text component where its current HTML structure causes a warning about HTML semantic issues. 

I have introduced a new prop called "as" that is passed to styled-components. [This prop is supported in styled-components since v4](https://styled-components.com/docs/api#as-polymorphic-prop), allowing us to specify which HTML element we want to use for the Skeleton.Text component. This way, we can use the Skeleton.Text component in a semantically correct way while maintaining flexibility.

With this PR the follow syntax's will be allowed

```tsx
{/* the default value for "as" is "div" */}
<Skeleton type="text" as="span" /> 
```
and 

```tsx
{/* the default value for "as" in Skeleton.Text is "span" because of semantic reasons */}
<Skeleton.Text /> 
```

Some refactors were made because this component was really asking for some care


components/Skeleton/Skeleton.jsx
```diff
@@ -1,85 +1,45 @@
...
   height: '14px',
   size: 'medium',
-  theme: {
-    baseFontSize: defaultBaseFontSize,
-    spacing,
-  },
 };
 
 Skeleton.propTypes = {
-  /** Sets the component behavior */
```
MarcosViniciusPC, line 65: These comments about the properties relate to their description when accessing the documentation (yarn storybook). Deleting them implies in making it difficult to understand the behavior of the components for those who are interested in using it.

components/Skeleton/__snapshots__/Skeleton.unit.test.jsx.snap
```diff
@@ -90,15 +94,15 @@ exports[`Skeleton component Should match the snapshot of a text skeleton 1`] = `
...
   width: 100%;
+  height: 14px;
 }
 
 .c1 {
   border-radius: 4px;
 }
 
-<div
+<span
```
MarcosViniciusPC, line 105: Why does only the Skeleton.Text snapshot render a span instead of a div? I didn't see changes in the unit tests that would justify this behavior, do you know what it could be?

components/Skeleton/Skeleton.jsx
```diff
@@ -1,85 +1,45 @@
...
   width: PropTypes.string,
-  /** Will affect only types that not uses the prop `size` */
   height: PropTypes.string,
-  /** Will affect only `Tag` and `Button` types.  */
   size: PropTypes.oneOf(['xsmall', 'small', 'medium', 'large']),
-  /** Used only for themification.  */
-  theme: PropTypes.shape({
-    spacing: PropTypes.object,
-    baseFontSize: PropTypes.number,
-  }),
```
MarcosViniciusPC, line 76: Why shouldn't we be able to send a custom theme to the component?

MarcosViniciusPC (changes requested):
I believe that defining a styles file separately instead of the current implementation of creating individual files inside a subcomponents folder is an architectural issue and changes in this regard should be discussed and coordinated in order to maintain the organization of the project and reduce the learning curve of new developers to the team, what do you think about it? Awaiting return.

components/Skeleton/Skeleton.jsx
2bfe26: It's not exactly that we shouldn't. Let me try to explain my reasoning for removing this property.

We currently use the theme only to derive other values, as shown in these examples:

Example 1:

```ts
  height: ${({ size }) => {
    const sizes = {
      small: baseFontSize * 0.875,
      medium: baseFontSize * 1.125,
      large: baseFontSize * 1.25,
    };

    return `${(sizes[size] || sizes.medium) + spacing.xxsmall * 2}px`;
  }};
```

Example 2:

```ts
 height: ${({ size }) => {
    const sizes = {
      xsmall: spacing.large,
      small: spacing.xlarge,
      medium: spacing.xxlarge,
      large: spacing.xxxlarge,
    };

    return `${sizes[size] || sizes.medium}px`;
  }};
```

In this situation, I believe that using the theme as a way to "extend" this component may not be the best approach. It could be challenging for users to effectively update these values using the theme. 

If you think it's worth keeping this prop, I'm open to updating the pull request accordingly.

2bfe26:
> I believe that defining a styles file separately instead of the current implementation of creating individual files inside a subcomponents folder is an architectural issue and changes in this regard should be discussed and coordinated in order to maintain the organization of the project and reduce the learning curve of new developers to the team, what do you think about it? Awaiting return.

I agree, it's a significant change in how components are defined, but it's a worthwhile change I think. It will make the components much easier to read. That been said, I'll update the pull request based on what you think its the best for the project.

components/Skeleton/__snapshots__/Skeleton.unit.test.jsx.snap
2bfe26: This pull request was created to address an issue (#437) related to semantic problems in the HTML structure of the Skeleton component. 

The suggestion in the issue was to use a `span` as the container element for `Skeleton.Text.` Therefore, in this pull request, I made the change to use a `span` as the **default element for Skeleton.Text**, as it was recommended to fix the semantic issue. 

There might not be any changes in the unit tests, as this is a structural change rather than a functionality change

components/Skeleton/Skeleton.jsx
2bfe26: Oh I was not aware of that. I will restore the comments. I apologize for the oversight.

components/Skeleton/Skeleton.jsx
```diff
@@ -1,85 +1,45 @@
...
-// Types
-Skeleton.Button = SkeletonTypes.Button;
-Skeleton.Circle = SkeletonTypes.Circle;
-Skeleton.Text = SkeletonTypes.Text;
-Skeleton.Tag = SkeletonTypes.Tag;
+Skeleton.Rect = props => <Skeleton {...props} />;
+Skeleton.Circle = props => <Skeleton type="circle" {...props} />;
+Skeleton.Text = props => <Skeleton type="text" as="span" {...props} />;
+Skeleton.Button = props => <Skeleton type="button" {...props} />;
+Skeleton.Tag = props => <Skeleton type="tag" {...props} />;
```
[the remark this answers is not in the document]
2bfe26: Honestly, I don't think it will cause any performance problems. The main concern here would be in terms of readability. Even if the user passes additional properties, it will not impact anything if those properties are not being used by the components.

In addition, I raised the idea of moving the codebase to TypeScript. With this migration we can create prop types for each of these variations, which would help to correct use by the user.

2bfe26:
These are thoughtful and fair comments. I have some disagreements with certain points, but I will update the pull request based on the final decision. 

Thank you for providing feedback.

components/Skeleton/__snapshots__/Skeleton.unit.test.jsx.snap
MarcosViniciusPC: Sorry I caused confusion on this one. I understand the purpose of the tweak, but what I didn't understand was why there is only one snapshot that renders a span instead of a div. I saw now that the Skeleton.Text defined in line 41 of the components/Skeleton/Skeleton.jsx file contained an "as" attribute defined with the value "span". Ignore this question, please.

components/Skeleton/Skeleton.jsx
MarcosViniciusPC: This issue involving the best use of theme as the default value for styling is already on our "radar". However, we didn't have a moment to think about effective solutions for this case. Regarding the last point mentioned by you, I believe that you removed the thema prop due to the overwriting of the default value of the props passed to the component by importing the thema and direct use in the component's styling, which consequently ends up overwriting the value of thema passed by the user and, with that, making its use unfeasible. However, I believe it would be interesting to keep the thema props and just adjust their use in order to avoid the previously mentioned problem, what do you think about that?

MarcosViniciusPC:
> > Acredito que definir um arquivo de estilos separadamente em vez da implementação atual de criar arquivos individuais dentro de uma pasta de subcomponentes é uma questão de arquitetura e mudanças nesse sentido devem ser discutidas e coordenadas para manter a organização do projeto e reduzir a curva de aprendizado de novos desenvolvedores para a equipe, o que você acha disso? Aguardo retorno.
> 
> Concordo, é uma mudança significativa na forma como os componentes são definidos, mas acho que vale a pena mudar. Isso tornará os componentes muito mais fáceis de ler.

MarcosViniciusPC:
> > I believe that defining a styles file separately instead of the current implementation of creating individual files inside a subcomponents folder is an architectural issue and changes in this regard should be discussed and coordinated in order to maintain the organization of the project and reduce the learning curve of new developers to the team, what do you think about it? Awaiting return.
> 
> I agree, it's a significant change in how components are defined, but it's a worthwhile change I think. It will make the components much easier to read.

Yes, but I think that coordinating actions in this sense helps to keep the project organized. I've seen cases of projects that involved several developers with different concepts of organization who implemented codes based on that without prior planning of actions and this made the project become a "salad of architecture and code structure" and that made it difficult the implementation of new functionalities to the project, because each new implementation followed a different organization. With that, I think we can combine thinking together about the best organization and implementing them in all components at once instead of doing this without planning actions and which components should be attacked, in order to maintain an implementation pattern , he understands?

Ah, sorry for closing the pr, I went to delete a previous comment and closed the pr by mistake. but already open again

components/Skeleton/Skeleton.jsx
2bfe26: I think it's understandable

2bfe26:
I would like to inform that the adjustments have been made 😇

MarcosViniciusPC (approved):
[x] Code review
[x] Unit
[x] Regression

rapahaeru (changes requested):
All good, but this branch needs rebase because the update of regression tests.
After that, will be approved.

2bfe26:
> All good, but this branch needs rebase because the update of regression tests. After that, will be approved.

done :)

MarcosViniciusPC (approved):

components/Skeleton/Skeleton.jsx
```diff
@@ -1,85 +1,84 @@
...
-import { spacing, baseFontSize as defaultBaseFontSize } from '../shared/theme';
-import {
-  SkeletonRect,
-  SkeletonCircle,
-  SkeletonText,
-  SkeletonButton,
-  SkeletonTag,
-} from './sub-components';
+import { theme as defaultTheme } from '../shared';
+import { Circle, Text, Button, Tag, Rect } from './sub-components';
```
rapahaeru, line 3: Nice!

rapahaeru (approved):

PR #505, opened 2023-06-13
fix(QTM-493): Updates component types

MarcosViniciusPC:
## Description
https://jirasoftware.catho.com.br/browse/QTM-493

## Review guide
- [ ] Code review
- [ ] TypeScript updated
- [ ] compatbile with React 18

### Before reviewing

Build the quantum and add it to the project with Typescript (preferably /vagas)
Remove all ts-ignore from the project Typescript.

### To review:

Create a page with all components and check one by one if intellisense works.

Check in the real components of the application if there are warnings of types being displayed. In this case, if there were typing problems in onClick functions, check if the callback passed in this prop contains the type required by the component or not.


components/Modal/index.d.ts
```diff
@@ -1,55 +1,61 @@
-import { ReactNode, TouchEvent, MouseEventHandler } from 'react';
+import { ReactNode, TouchEvent, MouseEventHandler, FC, HTMLAttributes } from 'react';
 
 export interface ModalProps {
   children?: ReactNode[] | ReactNode;
-  onClick?: (e?: MouseEvent | TouchEvent) => void;
+  onClick?: (e?: MouseEvent | TouchEvent) => void; // braking change
```
rapahaeru, line 5: why the comment?

MarcosViniciusPC: The default type of prop "e" (event) is MouseEvent only, TouchEvent is unnecessary. Thinking about removing this unnecessary type, it is possible that this generates breakchange as it may be that some application wrongly defines the type of the callback for the onClick prop

components/RadioGroup/index.d.ts
```diff
@@ -39,8 +38,6 @@ export interface RadioButtonProps {
   value: string;
 }
 
-export type RadioButton = ComponentType<RadioButtonProps>;
-
 export interface RadioGroupProps {
   type?: 'radio' | 'button';
   size?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge';
```
rapahaeru, line 43: onChange prop 

```
No overload matches this call.
  Overload 1 of 2, '(props: { type?: "button" | "radio" | undefined; name: string; defaultValue?: string | undefined; children?: ((string | number | boolean | {} | ReactElement<any, string | JSXElementConstructor<any>> | ReactNodeArray | ReactPortal | ReactNode[]) & (boolean | ... 2 more ... | ReactPortal)) | null | undefined; ... 5 more ...; required?: boolean | undefined; } & { ...; } & { ...; }): ReactElement<...>', gave the following error.
    Type '() => void' is not assignable to type '() => {} | ChangeEventHandler<HTMLInputElement>'.
      Type 'void' is not assignable to type '{} | ChangeEventHandler<HTMLInputElement>'.
  Overload 2 of 2, '(props: StyledComponentPropsWithAs<FC<RadioGroupProps> & { Radio: FC<RadioProps>; Button: FC<RadioButtonProps>; }, RadioGroupProps, {}, never, FC<...> & { ...; }, FC<...> & { ...; }>): ReactElement<...>', gave the following error.
    Type '() => void' is not assignable to type '() => {} | ChangeEventHandler<HTMLInputElement>'.ts(2769)
index.d.ts(51, 3): The expected type comes from property 'onChange' which is declared here on type 'IntrinsicAttributes & { type?: "button" | "radio" | undefined; name: string; defaultValue?: string | undefined; children?: ((string | number | boolean | {} | ReactElement<...> | ReactNodeArray | ReactPortal | ReactNode[]) & (boolean | ... 2 more ... | ReactPortal)) | null | undefined; ... 5 more ...; required?: bool...'
index.d.ts(51, 3): The expected type comes from property 'onChange' which is declared here on type 'IntrinsicAttributes & { type?: "button" | "radio" | undefined; name: string; defaultValue?: string | undefined; children?: ((string | number | boolean | {} | ReactElement<...> | ReactNodeArray | ReactPortal | ReactNode[]) & (boolean | ... 2 more ... | ReactPortal)) | null | undefined; ... 5 more ...; required?: bool...'
```

image

rapahaeru:
@MichelAraujo @rafael-marconato 
after the updates the react version to 18 on /vagas, the suggestions of the autocomplete stoped to work
even with the response coming correctly
image

same for "quero me candidatar" e "continuar lendo". Stoped to work.

image

components/Card/Thumbnail.d.ts
```diff
@@ -1,12 +1,14 @@
...
     src: string;
     alt: string;
     rounded?: boolean;
     theme?: { colors?: object };
-}> & {
-    displayName: 'Card.Thumbnail';
+}
+
+declare const Thumbnail: FC<ThumbnailProps> & {
+    displayName: "Card.Thumbnail";
```
souzaramon, line 11: Is it really necessary to declare displayName here?

components/SegmentedControl/index.d.ts
```diff
@@ -1,5 +1,7 @@
 import { FC, MouseEventHandler } from 'react';
 
+export type NonEmptyArray<T> = [T, ...T[]];
```
souzaramon, line 3: We should move this type to a location where it can be easily imported, to avoid duplication

components/Accordion/index.d.ts
```diff
@@ -1,4 +1,4 @@
-import { ReactNode, Component } from 'react';
+import { ReactNode, FC } from 'react';
 
 export interface AccordionProps {
     theme?: {
```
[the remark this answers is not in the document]
MarcosViniciusPC: I agree

components/Card/Thumbnail.d.ts
MarcosViniciusPC: I mistakenly thought that maybe it would help with debugging typescript applications, but I see that it doesn't. I will remove

components/Modal/index.d.ts
```diff
@@ -1,55 +1,61 @@
...
   onClose?: MouseEventHandler<HTMLButtonElement>;
   closeButtonAriaLabel?: string;
   theme?: {
-    breakpoints?: {};
-    colors?: {};
-    spacing?: {};
+    breakpoints?: object;
```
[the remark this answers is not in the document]
MarcosViniciusPC: I've been evaluating the project and changing the theme props type will take a long time and will extrapolate the priority of the task. I chose to keep it in this format and in the future adjust the theme types of all components

components/Dropdown/index.d.ts
```diff
@@ -1,4 +1,4 @@
-import { Component } from 'react';
+import { FC } from 'react';
 import { ControllerStateAndHelpers } from 'downshift';
 
 export type ItemPropType = string | { value: string | number; label: string };
```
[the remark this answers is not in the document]
MarcosViniciusPC: The callback passed to the onChange prop in this case contains arguments with different types than the implemented one. In this case, it is necessary to adjust the final application.

image
image

MarcosViniciusPC:
> textArea

the onBlur and onFocus methods of the TextArea need to be adjusted according to the quantum types

souzaramon (approved):
Looks good to me! This is a big improvement from what we had before. Most of the things we need to create a type now, will be inferred when we move the codebase to TypeScript. So I think it's okay to have some verbosity in the declaration files.

PR #511, opened 2023-07-05
Add false to error union at InputProps

ghost:
## Description
When using the [Formik](https://formik.org/), it is common to use the following pattern

```tsx
<Input
  name="holderName"
  label="Cardholder Name"
  error={formik.touched.holderName && formik.errors.holderName}
  onChange={formik.handleChange}
  onBlur={formik.handleBlur}
  onClean={onClean('holderName')}
/>
```
It is a very convenient way to pass the error. The expression **formik.touched.holderName && formik.errors.holderName** always returns a string or false. The component checks if the error is falsy and returns null in the error section if it is.

The only problem is when using the "strict" mode of TypeScript:
```Type 'string | false | undefined' is not assignable to type 'string | undefined'.```

I believe it's worth allowing the value "false" for these cases to work, but I don't think it makes sense to allow the Boolean type itself since the value "true" would likely be serialized and displayed on the screen.

### DIFF
--- a/components/Input/index.d.ts
+++ b/components/Input/index.d.ts
@@ -23,7 +23,7 @@ export interface InputProps extends InputHTMLAttributes<HTMLInputElement> {
   required?: boolean;
   placeholder?: string;
   type?: 'email' | 'text' | 'tel' | 'number' | 'password' | 'search';
-  error?: string;
+  error?: string | false;
   id?: string;
   name?: string;
   validate?: Validate | CustomValidate | Array<Validate | CustomValidate>;


MarcosViniciusPC (commented):
I saw that the adjustment made eliminates type warnings in VS Code but running the project I saw that in the console a warning is displayed when the prop 'error' is passed with value 'false'. Does this behavior occur for you? I think it has something to do with the PropType of the 'error' prop

image

2bfe26:
@MarcosViniciusPC I was not aware of this warning, so I'll investigate a little further.

PR #530, opened 2023-11-14
feat(QTM-511): Prop text of the Tooltip component is now of type node instead of string

MarcosViniciusPC:
## Description
https://jirasoftware.catho.com.br/browse/QTM-511

## Review guide
- [ ] Coverage
- [ ] Unit tests
- [ ] Doc
- [ ] Code review
- [ ] TypeScript updated

### DIFF
--- a/components/Tooltip/index.d.ts
+++ b/components/Tooltip/index.d.ts
@@ -1,7 +1,7 @@
 import { FC, ReactNode } from 'react';
 
 export interface TooltipProps {
-    text: string;
+    text: ReactNode;
     placement?: 'top' | 'right' | 'bottom' | 'left';
     visible?: boolean;
     multiline?: boolean;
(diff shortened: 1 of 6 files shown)


components/Tooltip/index.d.ts
alizeleal, line 1: Is this prop still limited to receive text?

I've tested with an image and it worked.

MarcosViniciusPC: Ideally, the component should receive text or textual elements (p, strong, a, etc...). I confess that I have doubts as to whether there is any scenario that will be used as an image. I'm going to ask @rapahaeru a question

MarcosViniciusPC: You can continue with the review @alizeleal, it's ok to keep the children receiving an image and having these size limitations, considering that the component's objective is to display small information

alizeleal (approved):
- [x]  Coverage
- [x]  Unit tests
- [x]  Doc

PR #532, opened 2023-12-19
fix(QTM-527): Fixing navigation between items using keyboard

MarcosViniciusPC:
## Description
https://jirasoftware.catho.com.br/browse/QTM-527

## Review guide
- [ ] Coverage
- [ ] Unit tests
- [ ] Code review

### A11y review
- [ ] A11y checks
  - [ ] accessible via keyboard?

### How to simulate the problem

Connect to VPN
Download the project https://github.com/catho/recruiter-experience_company-registration_app and use the latest version of the "main" branch.
Add an .env file to the project root with the correct data (ask @MarcosViniciusPC )
Install version 18 of node. Ex: `nvm install 18.16.1`
Change the node version to the previously installed version. Ex: `nvm use 18.16.1`
Download dependencies: yarn
Run the application: yarn dev
In the first step of registration, navigate using the tab to the "Job Role" field, type "desen" and select any suggested option
Wait 2 seconds (until the list of options appears again)
Press the "down" arrow key on the keyboard

### How to validate the fix

In quantum, install the dependencies and build: `yarn && yarn build`.
In the project https://github.com/catho/recruiter-experience_company-registration_app, add the local build of quantum. Ex: `yarn add ../quantum/dist`
Simulate the problem again
Carry out other navigation tests

### DIFF
--- a/components/AutoComplete/AutoComplete.jsx
+++ b/components/AutoComplete/AutoComplete.jsx
@@ -164,7 +164,7 @@ const AutoComplete = ({
     filterSuggestions.length,
   );
   const [showSuggestions, setShowSuggestions] = useState(false);
-  const [cursor, setCursor] = useState(0);
+  const [cursor, setCursor] = useState(-1);
   const wrapperRef = useRef();
   const listOptions = useRef();
   const autoInputRef = useRef(null);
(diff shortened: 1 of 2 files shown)


alizeleal (commented):
image
The bug seems to be fixed, but when loading the page, I'm getting this error message

MarcosViniciusPC:
> image The bug seems to be fixed, but when loading the page, I'm getting this error message

This error is specific to the application and not to any quantum component or catho-components. Some of the endpoints used in the env are not active and consequently errors occur in the requests. However, this is not a problem because even with these errors we were able to simulate the reported problem.

PR #576, opened 2024-08-01
feat(QTM-752):  Migrated SegmentedControl component and use of shared styles with the button

MarcosViniciusPC:
## Description
https://jirasoftware.catho.com.br/browse/QTM-752
https://jirasoftware.catho.com.br/browse/QTM-754

## Review guide
- [ ] Unit tests
- [ ] Regression
- [ ] Code review
- [ ] TypeScript updated
- [ ] Build

## Test Guide

This PR contains the following implementations

- Correction of the semantics of the SegmentedControl component, in which the button tag was changed to label, a standard adopted before the migration of the Button component to CSS Modules

To test, run the storybook and inspect the page, checking if the label attribute is displayed instead of the button

- Using shared styles between Button and SegmentedControl components

To test, do code review and check if a single css file is used to style the SegmentedButton and the Button

- Removing unused styles from the Alert component

To test, run the storybook and check if the component layout is the same as the version implemented in the master branch

- Correction of css build error caused when there were styles in nested folders (Ex: shared/styles)

To test, check if the /dist/index.css file created after the build contains the stylings from the /shared/styles/buttonbase.module.css file

- Correction of styling problems (:hover) of the Button.Icon component caused by the lack of migration of styles to CSS and by not using the size prop when defining the width. This last problem already existed even before the migration of the Button component to CSS.

To test, replace the const TemplateIconOnly in the file stories/Buttton/Button.regression-test.story.jsx to `const TemplateIconOnly = (args) => <Button.Icon icon="info" {...args} />` and run the backtest storybook and check the stories (IconOnly, IconOnlySmall, IconOnlyXSmall, etc.). In these stories, pass the skin prop as arg and see if there was a color change in the component in its ":hover" state


alizeleal (approved):
- [x]  Unit tests
- [x]  Regression
- [x]  Code review
- [x]  Build

PR #577, opened 2024-08-14
import Acute icon by material-icons

HallefHLVieira:
## Description
Implement an import to get Acute icon by material-icons


MarcosViniciusPC (changes requested):
[ ] Doc (yarn storybook)

The documentation (storybook) is broken, apparently because the new icon that was added was not included in the mock file. Here is an example pr to help you solve the problem. https://github.com/catho/quantum/pull/566

